PR #3, opened 2017-10-29
Reduce code smell

KengoTODA:
Reduce code smells which reported at [Sonarcloud](https://sonarcloud.io/component_measures?id=com.github.spotbugs.gradle&metric=code_smells)


KengoTODA:
@spotbugs/everyone Please check. I think the alert from sonarqube can be ignored, since they're not related with changes in this branch.

PR #6, opened 2018-02-09
Add option to control progress reporting

boidolr:
Fixes https://github.com/spotbugs/spotbugs/issues/506. Behaviour and naming is similar to the FindBugs plugin (see https://github.com/gradle/gradle/blob/master/subprojects/code-quality/src/main/groovy/org/gradle/api/plugins/quality/internal/findbugs/FindBugsSpecBuilder.java#L168).

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -6,6 +6,8 @@ Currently the versioning policy of this project follows [Semantic Versioning](ht
 
 ## Unreleased - 2017-??-??
 
+* Fixed "loud output" [#506](https://github.com/spotbugs/spotbugs/issues/506) by adding an option to show progress and disabling it by default, matching the behaviour of the FindBugs Gradle Plugin
+
 ## 1.6.0 - 2017-10-25
 
 * Use SpotBugs 3.1.0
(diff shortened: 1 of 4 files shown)


boidolr:
That will lead to the defaults being different from the findbugs plugin. Personally I consider the progress option being set by default here as a bug.
If you disagree I will change it though.

KengoTODA:
Got it. Then could you:
1. Update [CHANGELOG.md](https://github.com/spotbugs/spotbugs-gradle-plugin/blob/master/CHANGELOG.md) to add `## Fixed` after `Unreleased`, then explain this change. [this site](http://keepachangelog.com/en/1.0.0/) will help you about how we update CHANGELOG.
2. Update [version in build.gradle])(https://github.com/spotbugs/spotbugs-gradle-plugin/blob/master/build.gradle#L13) to `1.6.1-SNAPSHOT`

KengoTODA:
To pass build with Java9, I've proposed #7.

KengoTODA:
#7 has been merged. Could you rebase/merge latest master branch? Then build on Java9 should be stable. @boidolr

boidolr:
@KengoTODA done.

boidolr:
@KengoTODA I've missed to update the task mapping (see https://github.com/boidolr/spotbugs-gradle-plugin/commit/0ca39b66e09fd8a9d3be29446fe8927e4bde2dc1). To you wish to include that as well?

KengoTODA:
yes if necessary

PR #12, opened 2018-02-27
Handle multiple output source directories

KevinMcT:
* Modification to handle changes made in gradle 4.0
	- See https://docs.gradle.org/4.0/release-notes.html
	Location of classes in the build directory

### DIFF
--- a/src/main/java/com/github/spotbugs/SpotBugsPlugin.java
+++ b/src/main/java/com/github/spotbugs/SpotBugsPlugin.java
@@ -6,6 +6,7 @@ import java.io.InputStream;
 import java.io.UncheckedIOException;
 import java.net.URL;
 import java.util.Properties;
+import java.util.concurrent.Callable;
 import java.util.stream.StreamSupport;
 
 import org.gradle.api.InvalidUserDataException;
@@ -13,6 +14,7 @@ import org.gradle.api.Project;
 import org.gradle.api.artifacts.Configuration;
 import org.gradle.api.artifacts.DependencySet;
 import org.gradle.api.file.ConfigurableFileTree;
+import org.gradle.api.file.FileCollection;
 import org.gradle.api.internal.ConventionMapping;
 import org.gradle.api.plugins.quality.CodeQualityExtension;
 import org.gradle.api.plugins.quality.internal.AbstractCodeQualityPlugin;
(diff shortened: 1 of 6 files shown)


src/main/java/com/github/spotbugs/SpotBugsPlugin.java
```diff
@@ -245,15 +245,21 @@ protected void configureForSourceSet(final SourceSet sourceSet, SpotBugsTask tas
...
+                 * As a result of the changes made in gradle 4.0.
+                 * See https://docs.gradle.org/4.0/release-notes.html - Location of classes in the build directory
+                 * Compile no longer bundles all classes in one directory build-gradle/classes/main
+                 * but instead separates classes into build-gradle/classes/{language}/main.
+                 *
+                 * We must therefor retrieve all output directories. Filter away the once that don't exist. Add each
+                 * existing file tree dependency to specified task. And then return the complete fileCollection, contain
+                 * all .class files available for analysis.
+                 */
+                FileCollection presentClassDirs = sourceSet.getOutput().getClassesDirs().filter(File::exists);
```
KevinMcT: filtering prevents a NoSuchFileException from being throw down in fb/sb internals as it will attempt to access directories even though they don't exist. 
A better approach would be to handle this exception where it arises and skip this "potentially" costly operation here.

KevinMcT:
Yeah, I'll look into it when I get back from holidays.  :)

13. mar. 2018 02:07 skrev "Kengo TODA" <notifications@github.com>:

terribleherbst:
Any progress on this?

KevinMcT:
Forgot a bit about it, on paternity leave atm. Picking this up now. Will
update with test coverage latest thursday.

*----------------------------------------------------*
Vennlig hilsen
*Kevin Mc Tiernan*

e-post: kmctiernan@gmail.com
mobil: +47 47336615

KevinMcT:
@terribleherbst @KengoTODA having some trouble writing tests to replicate the failure scenario / success scenario with the fix.

The tests I've written pass succesfully regardles of fix or not, however when I copy the project created by the tests and run the tasks manually through console/IDE, I get the expected results.

Simple example of the test I wrote which should "fail". But spotbugsMain task reports Success.

`public class IssueTest {
    @Rule
    public TemporaryFolder folder= new TemporaryFolder();

    @Before
    public void createProject() throws IOException {
        String buildScript = "plugins {\n" +
                "  id 'java'\n" +
                "  id 'groovy'\n" +
                "  id 'com.github.spotbugs' version '1.6.1'\n" +
                "}\n" +
                "version = 1.0\n" +
                "repositories {\n" +
                "  mavenCentral()\n" +
                "  mavenLocal()\n" +
                "}\n" +
                "dependencies {\n" +
                "    compile 'org.codehaus.groovy:groovy-all:2.4.14'\n" +
                "}\n";
        File buildFile = folder.newFile("build.gradle");
        Files.write(buildFile.toPath(), buildScript.getBytes(StandardCharsets.UTF_8), StandardOpenOption.WRITE);

        File sourceDir = folder.newFolder("src", "main", "groovy");
        File to = new File(sourceDir, "Bar.groovy");
        File from = new File("src/test/java/com/github/spotbugs/Bar.groovy");
        Files.copy(from.toPath(), to.toPath(), StandardCopyOption.COPY_ATTRIBUTES);
    }

    @Test
    public void testShouldFail() throws Exception {
        BuildResult result = GradleRunner.create().withProjectDir(folder.getRoot())
                .withArguments(Arrays.asList("spotbugsMain")).withPluginClasspath().build();
        assertThat(result.task(":compileGroovy").getOutcome(), is(TaskOutcome.SUCCESS));
        assertThat(result.task(":classes").getOutcome(), is(TaskOutcome.SUCCESS));
        assertThat(result.task(":spotbugsMain").getOutcome(), is(TaskOutcome.NO_SOURCE));
    }
}`

Got some idea's ? :)

terribleherbst:
I guess the test case should be created in a different way, at least I have a different scenario. 

I do have the following structure:
```
src/main/java/Bar.java
src/main/groovy (empty)
src/test/java/ (empty)
src/test/groovy/BarTest.groovy
```

Could you try this?

gcwiak:
Any updates?

KevinMcT:
Yeah, back @ work now. Will see if I can sort this out over the next couple of days.

KevinMcT:
@fholger I found the issue. Much to the credit of your example.

The gradleRunner seems to ignore the fact that we're specifying a version for the plugin in the buildscript. As such is only uses the available local code. (Which sort of makes sense.) But on the other hand it ruins my initial take on the test.

So after conferring with one of my colleagues I think we've found a fair enough test proposal.

I'll push a second PR with only the test in it, which will fail. And update this PR with the test as well which will run the test successfully.

Given that proof, @KengoTODA I reckon you've got validation to submit this PR.

The 2nd failing PR can then be abandoned as there is no point point merging that in first producing an erroneous build just to fix it with a 2nd commit.

KevinMcT:
Bah fcked up a bit on the conflict resolution >.< Serves me right for getting so used to gerrit..

KengoTODA:
Sorry but we still have conflict, because we recently merged #3 and #18.

KevinMcT:
@KengoTODA https://github.com/spotbugs/spotbugs-gradle-plugin/pull/28 is running without the proposed fix. 

Hope this is good enough.

PS. Don't merge #28 💀  :)

KevinMcT:
@KengoTODA flow got a bit interrupted as I didn't use a feature branch for this PR. But requested changes have been added and conflicts resolved.

Good to merge? 😺

PR #28, opened 2018-05-23
Handle multiple output source directories - FAIL

KevinMcT:
** SHOULD NOT BE MERGED - ABANDON UPON MERGE OF: PULL#12 **
github.com/spotbugs/spotbugs-gradle-plugin/pull/12

Commit only contains the tests which verify that the fix contained within
pull request #12 fixes these issues.
These test will fail upon build. Check PR#12 to confirm that the fix
results in a succesful build.


KengoTODA:
Understood, thanks for your reproduction!

Issue #5, opened 2017-11-25
provide default locations for include/exclude filters


xenoterracide:
I noticed that the checkstlye plugin in gradle looks for a configuration file at `config/checkstyle/checkstyle.xml` would be nice if spotbugs looked for `config/spotbugs/{include,exclude}.xml` by default, and ignored it if the file isn't present

binkley:
Is this issue still open?  I looked at the Feb 26 comment referenced by @boidolr, which appears to provide default files.  Does this work?

jscancella:
Closing due to lack of response.

PR #36, opened 2018-06-16
Default filter locations

boidolr:
I _think_ these changes should take care of https://github.com/spotbugs/spotbugs-gradle-plugin/issues/5. I am not fluent with the extension API though, so please double check that indeed the issue is fine and fixed without problems.


KengoTODA:
I'll close this PR because the conflict is not resolved for years.
please recreate PR if necessary.

PR #50, opened 2018-09-07
Upgrade Gradle to 4.10

KengoTODA:
Gradle released 4.10, then use it to keep supporting latest version.


KengoTODA:
Build is failing because our gradle plugin depends on internal class in Gradle. We need a breaking change to support Gradle 5.0.

Issue #11, opened 2018-02-26
Process 'Gradle SpotBugs Worker 4' finished with non-zero exit value 1


hongqijia:
* What went wrong:
Execution failed for task ':Core:spotbugsMain'.
> Failed to run Gradle SpotBugs Worker
   > Process 'Gradle SpotBugs Worker 5' finished with non-zero exit value 1

* Try:
Run with --info or --debug option to get more log output. Run with --scan to get full insights.

* Exception is:
org.gradle.api.tasks.TaskExecutionException: Execution failed for task ':Core:spotbugsMain'.
        at org.gradle.api.internal.tasks.execution.ExecuteActionsTaskExecuter.executeActions(ExecuteActionsTaskExecuter.java:103)
        at org.gradle.api.internal.tasks.execution.ExecuteActionsTaskExecuter.execute(ExecuteActionsTaskExecuter.java:73)
        at org.gradle.api.internal.tasks.execution.OutputDirectoryCreatingTaskExecuter.execute(OutputDirectoryCreatingTaskExecuter.java:51)
        at org.gradle.api.internal.tasks.execution.SkipUpToDateTaskExecuter.execute(SkipUpToDateTaskExecuter.java:59)
        at org.gradle.api.internal.tasks.execution.ResolveTaskOutputCachingStateExecuter.execute(ResolveTaskOutputCachingStateExecuter.java:54)
        at org.gradle.api.internal.tasks.execution.ValidatingTaskExecuter.execute(ValidatingTaskExecuter.java:59)
        at org.gradle.api.internal.tasks.execution.SkipEmptySourceFilesTaskExecuter.execute(SkipEmptySourceFilesTaskExecuter.java:101)
        at org.gradle.api.internal.tasks.execution.FinalizeInputFilePropertiesTaskExecuter.execute(FinalizeInputFilePropertiesTaskExecuter.java:44)
        at org.gradle.api.internal.tasks.execution.CleanupStaleOutputsExecuter.execute(CleanupStaleOutputsExecuter.java:88)
        at org.gradle.api.internal.tasks.execution.ResolveTaskArtifactStateTaskExecuter.execute(ResolveTaskArtifactStateTaskExecuter.java:62)
        at org.gradle.api.internal.tasks.execution.SkipTaskWithNoActionsExecuter.execute(SkipTaskWithNoActionsExecuter.java:52)
[...]

PR #58, opened 2018-10-10
Fix #57. Implement Worker API support.

### LINKED ISSUE #57: Implement Worker API
Third-party plugins should leverage the Worker API instead of relying on Gradle internal APIs.

Worker API docs: https://guides.gradle.org/using-the-worker-api/

The only caveat I see so far is that this will break compatibility with Gradle versions earlier than 3.5.

DPUkyle:
- Remove dead code
 - Fix javadoc warning

Signed-off-by: Kyle Moore <github@kylemoore.com>


DPUkyle:
こちらこそ、お世話様です。よろしくお願いいたします。

DPUkyle:
This should also fix #11. Could the OP please try verifying once more?

Issue #16, opened 2018-03-28
No way to configure analysis properties


Raniz85:
I think [these](http://spotbugs.readthedocs.io/en/latest/analysisprops.html) should be passed as system properties, but there is no way of setting system properties on the _SpotBugsTask_ or on the _SpotBugsExtension_ so they can't be configured.

mannodermaus:
We'd like to include the `findbugs.sf.comment` exemption into our SpotBugs task, however the lack of support for system properties on the `SpotBugsTask` is a blocker here.

@Raniz85, did you ever figure out a workaround for this?

zbynek:
This should be possible with `jvmArgs` and `-D` to set a property.
```
spotbugs {
    jvmArgs = ['-Dfindbugs.sf.comment=true']
}
```

mannodermaus:
@zbynek That's true, this works since my linked PR was merged for 1.6.5!

PR #59, opened 2018-10-12
Keep track of top-level source set folders

mannodermaus:
SpotBugs’ source analysis requires a handle to the top-level source set
folder, or else the SourceFileFinder cannot locate attached sources
of a task. This is especially annoying when some checks require insight
into the source code (e.g. the SwitchFallthrough check, which looks
at comments inside a switch statement). Since SpotBugs tasks are essentially
Gradle SourceTasks, they use a FileTree internally, which only keeps
handles to the leaf files automatically, and doesn’t hold onto the original
directory:

```groovy
new File("src/main/java/com/example/MyFile.java").withWriter {
  it.write("package com.example;\nclass MyFile {}"
}

def fileTree = project.fileTree("src/main/java")
for (file in fileTree.files) {
  // This will not print "/.../src/main/java", but "/.../src/main/java/com/example/MyFile.java" directly.
  println file.absolutePath
}
```

This commit expands the previous behavior with the ability to obtain
the source set folder, and obtains a combined FileCollection to give
to SpotBugs through the new SpotBugsTask#getAllSource() method.

This resolves inquiries like https://github.com/spotbugs/spotbugs-gradle-plugin/issues/16, which are now supported through `jvmArgs` on either the `SpotBugsExtension` or its tasks:

```groovy
tasks.withType(com.github.spotbugs.SpotBugsTask) {
  jvmArgs = ['-Dfindbugs.sf.comment=true']
}
```

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -6,6 +6,8 @@ Currently the versioning policy of this project follows [Semantic Versioning](ht
 
 ## Unreleased - 2018-??-??
 
+* Keep track of top-level source set folders [#59](https://github.com/spotbugs/spotbugs-gradle-plugin/pull/59)
+
 ## 1.6.4 - 2018-09-26
 
 ### Changed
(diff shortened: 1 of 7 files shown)


mannodermaus:
Thanks for the review. I added a changelog entry!

PR #67, opened 2018-10-31
Use runtime classpath instead of compile classpath

### LINKED ISSUE #62: Hypothesis: should use getRuntimeClasspath instead of getCompileClasspath
In `configureForSourceSet`, the classpath is mapped as `taskMapping.map("classpath", sourceSet::getCompileClasspath);`
However, if the compilation depends on multiple layers of module dependencies that only expose their dependencies as `implementation`, those dependencies do not end up on the compile classpath, thus leading to `The following classes needed for analysis were missing:` messages.
I'm guessing it should be enough to replace `getCompileClasspath` with `getRuntimeClasspath` to rectify this.

KengoTODA:
this PR closes #62.

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -6,6 +6,10 @@ Currently the versioning policy of this project follows [Semantic Versioning](ht
 
 ## Unreleased - 2018-??-??
 
+### Fixed
+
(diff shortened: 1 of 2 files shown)


KengoTODA:
closed by #79

PR #79, opened 2018-12-17
Use runtime classpath instead of compile classpath

KengoTODA:
Same with #67 but rebased to the latest `master` branch.
It also fixes the wrong CHANGELOG entry introduced by #59.


KengoTODA:
@spotbugs/core-devs please review this change, thank you!

Issue #81, opened 2018-12-18
version 1.6.7 is not available at plugin site


KengoTODA:
We cannot publish gradle plugin version 1.6.7 to https://plugins.gradle.org/plugin/com.github.spotbugs due to following error:

```
Failed to post to server.
Server responded with:
Plugin: 'com.github.spotbugs', version: '1.6.7' exists already. It was created: 2017-05-16 13:43:18.304
```

But of course we have no 1.6.7 in plugin page. I'll ask Gradle community about why.

KengoTODA:
I posted a topic at [gradle forum](https://discuss.gradle.org/t/plugin-cannot-be-published-due-to-exists-already-error/29911).

KengoTODA:
1.6.8 has been successfully deployed.

PR #87, opened 2018-12-21
Release 1.6.8

KengoTODA:
To solve #81, it is [necessary to deploy newer version](https://discuss.gradle.org/t/plugin-cannot-be-published-due-to-exists-already-error/29911/2) so this PR is going to release new version 1.6.8.

It also bump up SpotBugs version to the latest one.


KengoTODA:
@spotbugs/core-devs please have a review, thank you :)

Issue #54, opened 2018-09-24
classes filtering does not work anymore


jkanto:
Since plugin version 1.6.2 it seems that classes filtering cannot be done. Example:
```
spotbugsTest {
	classes = classes.filter {
		!it.path.contains("foo/")
	}
}
```
With plugin version 1.6.1 classes on foo package are not processed, but since 1.6.2 filtering is ignored and all classes are processed by SpotBugs.

wrprice:
Workaround is to use an [exclude filter file](https://spotbugs.readthedocs.io/en/stable/filter.html), which supports excluding packages by regular expression.

PR #88, opened 2018-12-21
Fix classes filtering

KengoTODA:
In the closure for `classes`, previously `it` returns each class file, however now it just return the root dir (classesDirs). a2740fff2c58d05c7720f7e819daf0ca827b95cd introduced this problem #54.

By this fix, we can resume using closure based filtering.

### DIFF
--- a/src/test/java/com/github/spotbugs/Issue54Test.java
+++ b/src/test/java/com/github/spotbugs/Issue54Test.java
@@ -1,0 +1,79 @@
+/*
+ * Contributions to SpotBugs
+ * Copyright (C) 2017, kengo
+ *
+ * This library is free software; you can redistribute it and/or
+ * modify it under the terms of the GNU Lesser General Public
+ * License as published by the Free Software Foundation; either
+ * version 2.1 of the License, or (at your option) any later version.
+ *
+ * This library is distributed in the hope that it will be useful,
+ * but WITHOUT ANY WARRANTY; without even the implied warranty of
+ * MERCHANTABILITY or FITNESS FOR A PARTICULAR PURPOSE.  See the GNU
+ * Lesser General Public License for more details.
+ *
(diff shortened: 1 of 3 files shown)


KengoTODA:
@spotbugs/core-devs please review this change, thank you!

src/test/java/com/github/spotbugs/Issue54Test.java
```diff
@@ -0,0 +1,79 @@
...
+/**
+ * @see <a href="https://github.com/spotbugs/spotbugs-gradle-plugin/issues/54">GitHub issue</a>
+ */
+public class Issue54Test {
+    @Rule
+    public TemporaryFolder folder = new TemporaryFolder();
+
+    @Before
+    public void createProject() throws IOException {
+      String buildScript = "plugins {\n" +
```
ThrawnCA: This might be better as a filesystem resource, rather than concatenating strings. Consider fb-contrib's "CBX_CUSTOM_BUILT_XML" detector.

src/test/java/com/github/spotbugs/Issue54Test.java
```diff
@@ -0,0 +1,79 @@
+/*
+ * Contributions to SpotBugs
+ * Copyright (C) 2017, kengo
```
ThrawnCA: Should be 2018 since this is new.

src/test/java/com/github/spotbugs/Issue54Test.java
KengoTODA: Let me handle it in the separated PR, because we have many test cases that has Gradle config in source code.

src/test/java/com/github/spotbugs/Issue54Test.java
KengoTODA: Fixed.

PR #89, opened 2018-12-21
Fix warning reported by :validateTaskProperties task

KengoTODA:
When we build this project, `:validateTaskProperties` task reports one warning:

```
> Task :validateTaskProperties
Task property validation finished with warnings:
  - Warning: Task type 'com.github.spotbugs.SpotBugsTask': property 'allSource' has @Input annotation used on property of type org.gradle.api.file.FileCollection.
```

This change will fix this warning. I'm not sure we need to update CHANGELOG or not...


KengoTODA:
Duplicate of #96

PR #96, opened 2019-01-10
Make task cacheable across checkouts

manuranga:
allSource input creates different cache keys for same project checked out at different locations. This PR will fix it.


manuranga:
Thank you @KengoTODA.

PR #100, opened 2019-01-28
Revert 58 to fix spotbugs/spotbugs!819

KengoTODA:
This PR solves [a reported problem at the core project](https://github.com/spotbugs/spotbugs/issues/819).
To avoid potential issues, it also bumps up supported Gradle version from 4.0 to 5.0.


KengoTODA:
@spotbugs/everyone could you review these changes? thank you!

src/main/java/com/github/spotbugs/internal/spotbugs/SpotBugsExecuter.java
```diff
@@ -0,0 +1,37 @@
...
+
+import java.io.IOException;
+import java.util.List;
+
+import edu.umd.cs.findbugs.FindBugs;
+import edu.umd.cs.findbugs.FindBugs2;
+import edu.umd.cs.findbugs.IFindBugsEngine;
+import edu.umd.cs.findbugs.TextUICommandLine;
+
+public class SpotBugsExecuter implements SpotBugsWorker {
```
ThrawnCA: Misspelled "Executor"

build.gradle
```diff
@@ -14,6 +14,7 @@ apply from: "$rootDir/gradle/deploy.gradle"
 version = '1.6.10-SNAPSHOT'
 group = "com.github.spotbugs"
 def spotBugsVersion = '3.1.11'
+def slf4jVersion = '1.8.0-beta2'
```
ThrawnCA: I think we need to seriously consider dropping SLF4J and just using JUL, for maximum compatibility.

KengoTODA: If you mean the trouble around gradle plugin, I think all the reported issues can be solved if we stop depending on Gradle internal API that excludes SLF4J from the dependency. So I still think that SLF4J is better than that horrible JUL...

KengoTODA:
Updated, please review this PR again @ThrawnCA

oehme:
Why revert using the worker API? Looks like a bad idea to me, because now you depend on internal APIs again. Shouldn't instead the usage of static state be removed?

DPUkyle:
I agree with @oehme; using Gradle internal APIs is a step backwards.

If the spotbugs tool can not be quickly remedied to avoid preservation of static state, I suggest that we preserve the public API of `SpotBugsTask` while avoiding the Worker API or internal APIs. Instead, the `@TaskAction` method will invoke `Project#javaexec(Action)`; that spawned JVM will then die at the conclusion of the task execution. While this isn't ideal from a performance POV, it should prevent the static state issues identified by @Vampire.

KengoTODA:
I also believe that it's mandatory to remove dependency on Gradle internal API. This is why I made #104 recently.

At the same time, we need to make this product working. The reverted code in this PR [introduced a trouble](https://github.com/spotbugs/spotbugs/issues/819). To make it stable, I merged this PR after [the user test](https://github.com/spotbugs/spotbugs/issues/819#issuecomment-452641955).

Let's find the way to realize stable feature in both of feature and compatibility :)
I believe it's possible even though [SpotBugs itself is thread unsafe](https://github.com/spotbugs/spotbugs/issues/249#issuecomment-314432619).

KengoTODA:
[JavaExec](https://docs.gradle.org/current/dsl/org.gradle.api.tasks.JavaExec.html) seems nice. It will be nice option to solve this issue.

Vampire:
Maybe even a combination of worker api and javaExec could make sense.
Because by using the worker api, tasks are done in parallel up to max-workers amount.
The worker implementation could then spawn a separate process.
This way you have the parallelity and still the isolation.
With the current solution I think the tasks would again be done sequentially only, as would be by only doing a javaExec.

davidburstromspotify:
I must admit I haven't looked closely at the holistic issue, but what's problematic about dealing with the printout or dual registration at the point it happens? I don't see a dysfunction in the checkers where they themselves keep static state during analysis.

Vampire:
The problem is the same why in 80% of the cases when you do an `invokeLater` in Swing to fix a bug you even make it worse.
It is symptom treatment, not bug fixing and you just shift the problem to a less obvious and harder to solve place.
I don't know of whether there are other static things that are already broken in more subtle ways, but speaking of the plugins, imagine two projects A and B that both run with the same Gradle version and thus on the same daemon and the Spotbugs task will run on the same worker (e. g. wih `--max-workers 1`).
Project A uses the find-sec-bugs plugin, Project B does not.
Now you analyze project B and you get no bugs found.
Next you analyze project A, not warnings, no bugs found.
Now you analyze project B again without any change and (even without warning as there is no double registration) suddenly you get issues found.

Or you analyze A and B with different version of the same plugin, and so on.

davidburstromspotify:
@Vampire I totally agree with that assessment. What I'm meaning is that instead of printing a warning for every time a factory is re-registered, why not be pragmatic and warn only once? That doesn't hide the symptom, and it doesn't spam the logs, and you'll get the speedup associated with the worker API.
For sure, there could be cases where your isolation is broken. Is that an issue that people are currently struggling with, or is it more of a theoretical problem at this point?

Vampire:
What do you gain with speedup if you get wrong results? o_O

Besides that, if he follows my suggestion to combine worker-api and javaExec, you are just a bit slower as you need to start a VM each time, but it will still be done in parallel.

davidburstromspotify:
@Vampire If people _actually_ do get wrong results, it's not a gain, agreed. In our projects we run the same extensions for SpotBugs for all (hundreds of) modules, so the execution time problem is much more important to solve. The speedup with the worker API and in daemon processing was substantial, reducing the execution time with more than 75%. That being said, we can perform tests with your proposed solution and see if it performs well enough. If it is at least 50% faster than the baseline I'd consider it a success from our point of view, otherwise not.

Vampire:
It was not in-daemon with the worker API.
The isolation level was set to `PROCESS`, so the tasks are done in separate worker processes, not in the daemon process.
*Those* were reused though.
I'd guess the most speed gain was because of the parallel execution, so it might speed up enough for you with my proposal, but I might be wrong of course.

davidburstromspotify:
You're right, I was under the wrong impression with how the worker API was used.

I'd be happy to test the proposal, so we can know for sure.

build.gradle
swpalmer: +1 for JUL.. it just works.

ctmay4:
I'd like to add to what @davidburstromspotify said that the slowdown is a big deal for larger projects.  We had to roll back to a prior version (1.6.9) of the plugin because it made such a huge difference in our in CI builds.

KengoTODA:
Feel free to share your idea to solve this problem via PR! :)

KengoTODA:
note: It is possible to launch SpotBugs by `project.javaexec`, however it cannot create `SpotBugsResult` instance because we cannot touch `IFindBugsEngine` instance after execution.

```java
      ExecResult result = project.javaexec(new Action<JavaExecSpec>() {
		@Override
		public void execute(JavaExecSpec javaSpec) {
			javaSpec.args(spec.getArguments());
			javaSpec.setMain("edu.umd.cs.findbugs.LaunchAppropriateUI");
		}
      });
```

To keep current feature, we need to avoid using `JavaExec`, then probably need to hack a classloader?

Issue #101, opened 2019-02-07
Gradle 4.7 or higher is required


DPUkyle:
I made a mistake in implementing #66, as the method it calls is introduced in Gradle 4.7. This means that the SpotBugsTask in plugin 1.6.6 and higher will fail at runtime with Gradle 4.0~4.6.:

```
Execution failed for task ':foo:spotbugsMain'.
> com.github.spotbugs.internal.SpotBugsReportsInternal.getEnabledReports()Ljava/util/Map;

* Try:
Run with --debug option to get more log output. Run with --scan to get full insights.

* Exception is:
org.gradle.api.tasks.TaskExecutionException: Execution failed for task ':foo:spotbugsMain'.
...
Caused by: java.lang.NoSuchMethodError: com.github.spotbugs.internal.SpotBugsReportsInternal.getEnabledReports()Ljava/util/Map;
        at com.github.spotbugs.SpotBugsTask.run(SpotBugsTask.java:246)

```

I will fix the bug and add a regression test that will catch future API gotchas like this.

remal:
Please release it.

KengoTODA:
@remal 1.6.11 and 1.6.10 contain this change.

PR #102, opened 2019-02-07
Add regression test for #101 and fix bug with API-safe call

DPUkyle:
By inspecting the `GradleVersion`, we can avoid calling a method which might not exist at runtime.

Hopefully new test `SpotBugsPluginTest#testSpotBugsTaskCanRunWithMinimumSupportedVersion` can catch mistakes like this in the future.


DPUkyle:
Everything is passing now; let me know if you prefer I squash to a single commit before merging.

PR #117, opened 2019-02-18
Verify and fix code format automatically by Spotless

KengoTODA:
By this PR, we verify code automatically in CI. For instance, [this build](https://travis-ci.org/spotbugs/spotbugs-gradle-plugin/builds/494690094) was failed because it has wrongly formatted code. Then reviewer can focus not on code format but on important things.

Spotless also provides automatic fix, we can run `./gradlew spotlessApply` then it will fix all problems in the project.

### DIFF
--- a/gradle/spotless.gradle
+++ b/gradle/spotless.gradle
@@ -1,0 +1,7 @@
+spotless {
+  java {
+    removeUnusedImports()
+    trimTrailingWhitespace()
+    indentWithSpaces()
+  }
+}
--- a/build.gradle
+++ b/build.gradle
@@ -3,6 +3,7 @@ plugins{
   id 'java-gradle-plugin'
   id "com.gradle.plugin-publish" version "0.9.7"
   id 'org.sonarqube' version '2.6.2'
+  id 'com.diffplug.gradle.spotless' version '3.18.0'
 }
 
 apply from: "$rootDir/gradle/checkstyle.gradle"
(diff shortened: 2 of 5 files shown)


gradle/spotless.gradle
```diff
@@ -0,0 +1,7 @@
+spotless {
+  java {
+    removeUnusedImports()
```
wreulicke: why do not use any format such as eclipseFormatter or googleJavaFormat?

https://github.com/diffplug/spotless/blob/master/plugin-gradle/README.md#applying-to-java-source-with-google-java-format

wreulicke:
This is just a question.

Does this PR introduce format lint with spotless on travisci?

KengoTODA:
> Does this PR introduce format lint with spotless on travisci?

`spotlessCheck` task depends on `check` task, so it will run in the CI build:
https://github.com/spotbugs/spotbugs-gradle-plugin/blob/84b55ebfd723bdfb6898be5cacc372499b8ff64b/.travis.yml#L42

However `spotlessApply` does not run in CI for now.

gradle/spotless.gradle
KengoTODA: It's also possible solution, I think. I just wanted to keep the diff simple.

@spotbugs/everyone if you don't care, I will introduce the latest googleJavaFormat.

ThrawnCA:
Is this going to reformat the whole codebase at once?

Won't that mean a huge whitespace change in the commit history?

KengoTODA:
@ThrawnCA if we apply google java format, it is possible. Current change set doesn't contain big change, because we apply minimal format rules only.

KengoTODA:
Hello, can we merge this PR? Then please approve.

I think we don't need to add more rules in this timing.

PR #118, opened 2019-02-18
Remove no-op stream code

### LINKED ISSUE #47: Convention mapping for "classes" contains dead code
https://github.com/spotbugs/spotbugs-gradle-plugin/blob/63d4f0e3bf2044cb8e10c187d3b63de94e5aeed5/src/main/java/com/github/spotbugs/SpotBugsPlugin.java#L187

The whole "stream" code looks like a no-op to me, because its result is ignored. It can probably be removed, along with the whole comment above it. The mapping should just be "classes = sourceSet.output.classesDirs". This already carries all the necessary dependency information. 

Missing files can be handled later if the task actually executes. This saves some unnecessary IO if nothing has changed.

KengoTODA:
This PR fixes #47. I think it needs no CHANGELOG entry because it doesn't affect any output from this plugin.


KengoTODA:
@spotbugs/everyone please review this change. Thanks in advance!

PR #127, opened 2019-03-11
feat: Improvements for HTML reporting

### LINKED ISSUE #107: shipped html report stylesheets
SpotBugs ships a handfull of html stylesheets like for example `fancy-hist.xsl`.
The Gradle plugin only supports text resources, so to use them (without copying them into your project) you need to do something like
```gradle
configurations { spotbugsStylesheets { transitive false } }
dependencies { spotbugsStylesheets 'com.github.spotbugs:spotbugs:3.1.10' }
tasks.withType(SpotBugsTask) {
    reports {
        xml.enabled false
        html {
            enabled true
            stylesheet resources.text.fromArchiveEntry(configurations.spotbugsStylesheets, 'fancy-hist.xsl')
        }
    }
}
```
which is a bit cumbersome for using the stylesheets that are actually already shipped.
Would be nice if there were a `stylesheet` method that accepts the shipped stylesheets and transforms them to Gradle text resources automatically.

### LINKED ISSUE #111: -projectName and -release
Please by default set the options `-projectName` and `-release` to improve generated reports.

To do it manually you can for example do
```gradle
tasks.withType(SpotBugsTask) {
    extraArgs '-projectName', "$project.name ($it.name)"
    extraArgs '-release', "$project.version"
}
```

Effect is for example
grafik
vs.
grafik

KengoTODA:
Support `-projectName` and `-release` option that makes reporting better. By default, it uses project name and version.

Also add `setStylesheet(String)` for HTML report: For now we only has `setStylesheet(TextResource)` but XSL is packaged into `spotbugs.jar` then it is good to support the path for zip entries.

This PR closes #111. Also close #107.

### DIFF
--- a/src/main/java/com/github/spotbugs/internal/SpotBugsHtmlReportImpl.java
+++ b/src/main/java/com/github/spotbugs/internal/SpotBugsHtmlReportImpl.java
@@ -16,9 +16,9 @@ import org.gradle.api.tasks.Input;
 
 public class SpotBugsHtmlReportImpl extends CustomizableHtmlReportImpl {
     private static final long serialVersionUID = 6474874842199703745L;
-    private transient final ResourceHandler handler;
-    private transient final Configuration configuration;
-    private transient final Logger logger;
+    private final transient ResourceHandler handler;
+    private final transient Configuration configuration;
+    private final transient Logger logger;
 
     /**
      * Null-able string representing relative file path of XSL packaged in spotbugs.jar.
--- a/src/main/java/com/github/spotbugs/internal/SpotBugsHtmlReportImpl.java
+++ b/src/main/java/com/github/spotbugs/internal/SpotBugsHtmlReportImpl.java
@@ -16,9 +16,9 @@ import org.gradle.api.tasks.Input;
 
 public class SpotBugsHtmlReportImpl extends CustomizableHtmlReportImpl {
     private static final long serialVersionUID = 6474874842199703745L;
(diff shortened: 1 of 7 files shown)


KengoTODA:
@spotbugs/everyone please review this change. Thank you!

src/main/java/com/github/spotbugs/internal/SpotBugsHtmlReportImpl.java
```diff
@@ -0,0 +1,60 @@
...
+        }
+
+        TextResourceFactory factory = handler.getText();
+        Optional<File> spotbugs = configuration.files(this::find).stream().findFirst();
+        if (spotbugs.isPresent()) {
+            File jar = spotbugs.get();
+            logger.debug("Specified stylesheet ({}) found in spotbugs configuration: {}", stylesheet, jar.getAbsolutePath());
+            return factory.fromArchiveEntry(jar, stylesheet);
+        } else {
+            throw new InvalidUserDataException("Specified stylesheet (" + stylesheet + ") does not found in spotbugs configuration");
```
Vampire: "does" -> "was"

KengoTODA: thanks, issued https://github.com/spotbugs/spotbugs-gradle-plugin/pull/130

KengoTODA:
I found that this feature doesn't work with `spotbugs :4.0.0-beta1`. I will try to fix.

PR #137, opened 2019-04-10
 Use Gradle 5.3.1

rpalcolea:

PR #138, opened 2019-04-12
ci: skip sonarqube analysis when PR comes from forks

KengoTODA:
The secure variable is not used in the PR comes from forks.
So before we trigger SonarQube analysis, make sure the `$SONAR_LOGIN` variable exists or not.

This fix will solve the build failure in #137.


KengoTODA:
@spotbugs/everyone this PR is ready to merge. Please have a review :)

PR #139, opened 2019-04-15
WIP: use TRAVIS_SECURE_ENV_VARS instead to judge condition to run Sonar

KengoTODA:
This PR is made from the branch in the different repo. It should not run the SonarQube analysis.


KengoTODA:
confirmed that `fork` works as expected. I will merge #140 when oraclejdk related build becomes stable in Travis CI.

PR #150, opened 2019-05-30
Allow projects without a SourceSet (Eg. Android) to use this SpotBugs…

abishek3876:
…Task

If a project does not contain a sourceset, the execution of type SpotBugsTask fails with `path may not be null or empty string. path='null'` error.
This is because the sourceset is passed as an `@InputFiles` and it is null.

### DIFF
--- a/src/main/java/com/github/spotbugs/SpotBugsTask.java
+++ b/src/main/java/com/github/spotbugs/SpotBugsTask.java
@@ -5,6 +5,7 @@ import java.io.IOException;
 import java.util.ArrayList;
 import java.util.Arrays;
 import java.util.Collection;
+import java.util.Collections;
 import java.util.Set;
 import java.util.stream.Collectors;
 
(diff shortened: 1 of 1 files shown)


abishek3876:
Sure... Will try to add them up over the weekend.

rompic:
is there any update for this one? thanks!

KengoTODA:
version 4 should work even for Android project with plugin version 3.5, please have a try.

but it seems that Android plugin 3.6 removed task so this plugin doesn't work: #194
if you have idea to support both 3.5 and 3.6, please let us know.

davidburstrom:
The Android Gradle Plugin version is available as a static field in the JAR. This is how the Android Pitest plugin does it: https://github.com/koral--/gradle-pitest-plugin/blob/master/src/main/groovy/pl/droidsonroids/gradle/pitest/PitestPlugin.groovy#L37

PR #159, opened 2019-09-25
Update plugin to build with Gradle 6.x

rpalcolea:
Hi!

Few things for Gradle 6 support:

* Upgrade `com.diffplug.gradle.spotless` to `3.24.2`
   * Older versions fail in Gradle 6.x with the following: 
   > Execution failed for task ':spotlessJava'. You must declare outputs or use `TaskOutputs.upToDateWhen()` when using the incremental task API

* Remove `@Optional` from `showProgress` in `SpotBugsTask`.
   * Reason:  @Input properties with primitive type 'boolean' cannot be @Optional.

* Replace `compile` and `testCompile` with `implementation` and `testImplementation` as the former will be removed in Gradle 7.0 and now they trigger warnings on builds

### DIFF
--- a/build.gradle
+++ b/build.gradle
@@ -3,7 +3,7 @@ plugins{
   id 'java-gradle-plugin'
   id "com.gradle.plugin-publish" version "0.10.1"
   id 'org.sonarqube' version '2.7.1'
-  id 'com.diffplug.gradle.spotless' version '3.24.1'
+  id 'com.diffplug.gradle.spotless' version '3.24.2'
 }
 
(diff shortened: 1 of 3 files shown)


KengoTODA:
> Remove @Optional from showProgress in SpotBugsTask.

Will it affect the behaviour of this plugin? I wish it won't (because `boolean` cannot be null so actually it's already not optional).

rpalcolea:
Correct, it won’t as it is really non optional already

hli37:
Hi, I am wondering how to remove @Optional from showProgress in SpotBugsTask

PR #162, opened 2019-09-29
Replace with maven-publish plugin

KengoTODA:
Replace `maven` plugin with `maven-publish` plugin that will work even with Gradle 7.x. Currently we do not use Maven Central for this project, then we can also remove the usage of `signing` plugin.

Refs #159


KengoTODA:
hi @spotbugs/everyone, could you review this PR? Thanks in advance!

PR #167, opened 2019-10-12
Validate when plugin is configured

KengoTODA:
Current way create task instances even though it is not necessary.
This PR applies the same way with [the official CodeNarc plugin validating configuration in setter method](https://github.com/gradle/gradle/blob/df7ac484a2facbb4503ba911666367de22b1d16a/subprojects/code-quality/src/main/groovy/org/gradle/api/plugins/quality/CodeNarcExtension.java).

This PR also changes the type of Exception from `IllegalStateException` to `IllegalUserDataException` that is recommended to make problem clear. This is also the same way with the plugin in above.

@daanschipper can I ask you to confirm this fix?

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -1,3 +1,4 @@
+
 # Changelog
 
 This is the changelog for SpotBugs Gradle Plugin. This follows [Keep a Changelog v1.0.0](http://keepachangelog.com/en/1.0.0/).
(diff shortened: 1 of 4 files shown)


KengoTODA:
@spotbugs/everyone all builds has been passed successfully. Please have a review. Thanks in advance!

daanschipper:
I can confirm that the tasks are not created anymore during configuration time. This is a huge performance increase for our project! Would it be possible to create a new release as well after this is merged?

PR #178, opened 2019-11-29
Remove @Input from setStylesheet as it is deprecated behavior

rpalcolea:
This is not a priority, found this when using spotbugs lately:

```
 Task :spotbugsMain
Type 'SpotBugsHtmlReportImpl': setter method 'setStylesheet()' should not be annotated with: @Input. This behaviour has been deprecated and is scheduled to be removed in Gradle 7.0.
	at java.lang.Iterable.forEach(Iterable.java:75)
	at org.gradle.internal.execution.steps.ValidateStep.execute(ValidateStep.java:56)
	at org.gradle.internal.execution.steps.SkipEmptyWorkStep.lambda$execute$2(SkipEmptyWorkStep.java:78)
```


KengoTODA:
I'm not sure why it is possible & necessary, will try to find related documents in the Gradle community.

PR #199, opened 2020-03-04
Fix the issue 196

### LINKED ISSUE #196: 4.0.0 no longer included in "check"
If I update the plugin version from 3.0.0 to 4.0.0, `spotbugMain` and `spotBugsTest` no longer run as part of the `check` task.  They work if I run the tasks manually, but they used to be included as a verification step.  I looked through the changelog and did not see anything.  Was this intentional?

KengoTODA:
close #196

### DIFF
--- a/src/main/groovy/com/github/spotbugs/snom/SpotBugsPlugin.java
+++ b/src/main/groovy/com/github/spotbugs/snom/SpotBugsPlugin.java
@@ -27,7 +27,7 @@ public class SpotBugsPlugin implements Plugin<Project> {
   public static final String SLF4J_CONFIG_NAME = "spotbugsSlf4j";
   public static final String EXTENSION_NAME = "spotbugs";
 
-  private final Logger log = LoggerFactory.getLogger(SpotBugsPlugin.class);
+  private final Logger log = LoggerFactory.getLogger(getClass());
 
   @Override
   public void apply(Project project) {
--- a/src/main/groovy/com/github/spotbugs/snom/SpotBugsPlugin.java
+++ b/src/main/groovy/com/github/spotbugs/snom/SpotBugsPlugin.java
@@ -18,12 +18,16 @@ import edu.umd.cs.findbugs.annotations.Nullable;
 import org.gradle.api.Plugin;
 import org.gradle.api.Project;
 import org.gradle.api.Task;
+import org.slf4j.Logger;
+import org.slf4j.LoggerFactory;
 
 public class SpotBugsPlugin implements Plugin<Project> {
   public static final String CONFIG_NAME = "spotbugs";
   public static final String PLUGINS_CONFIG_NAME = "spotbugsPlugins";
   public static final String SLF4J_CONFIG_NAME = "spotbugsSlf4j";
   public static final String EXTENSION_NAME = "spotbugs";
+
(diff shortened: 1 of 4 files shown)


src/main/groovy/com/github/spotbugs/snom/SpotBugsPlugin.java
```diff
@@ -18,13 +18,17 @@
...
+import org.slf4j.Logger;
+import org.slf4j.LoggerFactory;
 
 public class SpotBugsPlugin implements Plugin<Project> {
   public static final String CONFIG_NAME = "spotbugs";
   public static final String PLUGINS_CONFIG_NAME = "spotbugsPlugins";
   public static final String SLF4J_CONFIG_NAME = "spotbugsSlf4j";
   public static final String EXTENSION_NAME = "spotbugs";
 
+  private final Logger log = LoggerFactory.getLogger(SpotBugsPlugin.class);
```
ThrawnCA: This could instead be `LoggerFactory.getLogger(MethodHandles.lookup().lookupClass());` so that it can be copy-pasted to other classes. Also, why not `static`?

KengoTODA: I'm not so positive to make copy-paste friendly in this case. The code  with `MethodHandles` isn't so intentional.

To avoid the pointed mistake, since we use JDK11 to build in CI, I wanna use Errorprone with [SLF4J plugin](https://github.com/KengoTODA/errorprone-slf4j) that can detect it. I'll make separated PR.

> Also, why not static?

It simply has no merit nor demerit. `Plugin` instance is created just few times per build.

KengoTODA: Ouch, we code in Groovy so Errorprone doesn't work. Now it's time to use SpotBugs! :)

ThrawnCA: Note that PMD will flag non-static loggers as a mistake unless passed into the constructor: https://pmd.github.io/latest/pmd_rules_java_errorprone.html#properlogger

Why not make logging statements copy-paste friendly? They're gonna be copy-pasted at some point, may as well avoid the risk of forgetting to update the classname.

KengoTODA: Confirmed that SpotBugs SLF4J plugin cannot detect the issue, because Groovy compiles this code into the following format:

```
        Object var5 = var4[0].call(LoggerFactory.class, Object.class);
        this.log = (Logger)ScriptBytecodeAdapter.castToType(var5, Logger.class);
```

so I'll follow your suggestion at this moment.

KengoTODA: > Why not make logging statements copy-paste friendly?

I believe it's not the way to go. And if I do, I will choose `LoggerFactory.getLogger(getClass())` which is more intuitive code to read.

BTW static-or-not is not so problem, [FAQ says both are OK](http://www.slf4j.org/faq.html#declared_static). Then I personally prefer the less-magic way.

ThrawnCA: `getClass()` works for an instance logger, yes.

PR #204, opened 2020-03-09
Try to reproduce the reported case, and fix it

### LINKED ISSUE #196: 4.0.0 no longer included in "check"

KengoTODA:
This fix makes SpotBugs plugin makes the dependency onto `check` task when `java-base` plugin is applied. This change close #196 in the version `4.0.1` which expects that `java-base` plugin is already applied to the project.


KengoTODA:
@spotbugs/everyone please review this PR, thanks in advance!

PR #234, opened 2020-04-24
fix: Use task creation avoidance api

daanschipper:
Using `getTasks().register()` avoids creating the tasks during configuration time.
Currently the plugin is creating  number of modules * number of sourcesets amount of tasks, slowing down the configuration time.

Tasks created during configuration time, from a build scan

How can I build the plugin to test it locally to confirm that this solves the issue?

### DIFF
--- a/src/main/groovy/com/github/spotbugs/snom/internal/SpotBugsTaskFactory.java
+++ b/src/main/groovy/com/github/spotbugs/snom/internal/SpotBugsTaskFactory.java
@@ -47,7 +47,7 @@ public class SpotBugsTaskFactory {
                         log.debug("Creating SpotBugsTask for {}", sourceSet);
                         project
                             .getTasks()
-                            .create(
+                            .register(
                                 name,
                                 SpotBugsTask.class,
                                 task -> {


KengoTODA:
Confirmed that this fix can pass existing test cases, then #196 won't be reproduced by this change. This means that #204 is the necessary fix and #199 isn't, actually. 

Thank you!

> How can I build the plugin to test it locally to confirm that this solves the issue?

The suggested way is to make integration tests instead of the manual checks.
It is also possible to [install the artifact to the local maven repository](https://github.com/spotbugs/spotbugs-gradle-plugin/blob/4.0.5/.github/CONTRIBUTING.md).

For this PR, it seems safe to merge, then I'll merge this for performance enhancements.

KengoTODA:
semantic-release didn't trigger a release, because I forgot to use 'squash and merge' which uses PR name as commit comment, instead of 'rebase and merge'. I will release the next patch version later.

https://github.com/spotbugs/spotbugs-gradle-plugin/runs/619575381#step:8:55

KengoTODA:
v4.0.6 contains this fix.

https://github.com/spotbugs/spotbugs-gradle-plugin/releases/tag/4.0.6

PR #237, opened 2020-04-29
fix: use task creation avoidance api

daanschipper:
The issue of tasks being created during the configuration time still persists, the fix in #234 was not sufficient.

Any help with writing the functional test to cover this is greatly appreciated, I cannot find how to assert that a task has not been initialized when calling `./gradlew help`

### DIFF
--- a/src/main/groovy/com/github/spotbugs/snom/SpotBugsPlugin.java
+++ b/src/main/groovy/com/github/spotbugs/snom/SpotBugsPlugin.java
@@ -16,7 +16,7 @@ package com.github.spotbugs.snom;
 import com.github.spotbugs.snom.internal.SpotBugsTaskFactory;
 import org.gradle.api.Plugin;
 import org.gradle.api.Project;
-import org.gradle.api.Task;
+import org.gradle.api.plugins.JavaBasePlugin;
 import org.slf4j.Logger;
 import org.slf4j.LoggerFactory;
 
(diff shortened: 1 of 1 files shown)


src/main/groovy/com/github/spotbugs/snom/SpotBugsPlugin.java
```diff
@@ -39,8 +39,8 @@ public void apply(Project project) {
             javaBase -> {
               log.debug(
                   "The javaBase plugin has been applied, so making the check task depending on all of SpotBugsTask");
-              Task check = project.getTasks().getByName("check");
-              project.getTasks().withType(SpotBugsTask.class, check::dependsOn);
+              project.getTasks().named(JavaBasePlugin.CHECK_TASK_NAME).configure(task ->
```
KengoTODA, line 42: At this moment, we're sure that the `java-base` plugin has been applied to the project.  
Do we still need to defer the task creation?

daanschipper: With `project.getTasks().getByName()` the `check` task is still created, see https://docs.gradle.org/current/userguide/task_configuration_avoidance.html for more info.

KengoTODA: ah I got the point. even though the plugin has been applied, it's still possible that user don't need it so we may skip creating the check task for better performance. thanks!

Issue #239, opened 2020-05-08
Argument list too long issue


rpalcolea:
Hi folks,

Found that having projects with massive classpaths can not be used with spotbugs plugin. For example:

```
17:29:09 Caused by: org.gradle.process.internal.ExecException: Process 'command '/usr/lib/jvm/zulu-8-amd64/bin/java'' could not be started because the command line exceed operating system limits.
17:29:09 	at org.gradle.process.internal.DefaultExecHandle.execExceptionFor(DefaultExecHandle.java:241)
17:29:09 	at org.gradle.process.internal.DefaultExecHandle.setEndStateInfo(DefaultExecHandle.java:218)
17:29:09 	at org.gradle.process.internal.DefaultExecHandle.failed(DefaultExecHandle.java:372)
17:29:09 	at org.gradle.process.internal.ExecHandleRunner.run(ExecHandleRunner.java:87)
17:29:09 	at org.gradle.internal.operations.CurrentBuildOperationPreservingRunnable.run(CurrentBuildOperationPreservingRunnable.java:42)
17:29:09 	... 3 more
17:29:09 Caused by: net.rubygrapefruit.platform.NativeException: Could not start '/usr/lib/jvm/zulu-8-amd64/bin/java'
17:29:09 	at net.rubygrapefruit.platform.internal.DefaultProcessLauncher.start(DefaultProcessLauncher.java:27)
17:29:09 	at net.rubygrapefruit.platform.internal.WrapperProcessLauncher.start(WrapperProcessLauncher.java:36)
17:29:09 	at org.gradle.process.internal.ExecHandleRunner.startProcess(ExecHandleRunner.java:98)
17:29:09 	at org.gradle.process.internal.ExecHandleRunner.run(ExecHandleRunner.java:71)
17:29:09 	... 4 more
17:29:09 Caused by: java.io.IOException: Cannot run program "/usr/lib/jvm/zulu-8-amd64/bin/java" (in directory "/agent/workspace/myproject"): error=7, Argument list too long
17:29:09 	at net.rubygrapefruit.platform.internal.DefaultProcessLauncher.start(DefaultProcessLauncher.java:25)
17:29:09 	... 7 more
17:29:09 Caused by: java.io.IOException: error=7, Argument list too long
```

[...]

PR #243, opened 2020-05-08
feat: add support for auxclasspathFromFile in java process

rpalcolea:
Hi @KengoTODA 

This is to address https://github.com/spotbugs/spotbugs-gradle-plugin/issues/239

Found that having projects with massive classpaths can not be used with spotbugs plugin. For example:

```
17:29:09 Caused by: org.gradle.process.internal.ExecException: Process 'command '/usr/lib/jvm/zulu-8-amd64/bin/java'' could not be started because the command line exceed operating system limits.
17:29:09 	at org.gradle.process.internal.DefaultExecHandle.execExceptionFor(DefaultExecHandle.java:241)
17:29:09 	at org.gradle.process.internal.DefaultExecHandle.setEndStateInfo(DefaultExecHandle.java:218)
17:29:09 	at org.gradle.process.internal.DefaultExecHandle.failed(DefaultExecHandle.java:372)
17:29:09 	at org.gradle.process.internal.ExecHandleRunner.run(ExecHandleRunner.java:87)
17:29:09 	at org.gradle.internal.operations.CurrentBuildOperationPreservingRunnable.run(CurrentBuildOperationPreservingRunnable.java:42)
17:29:09 	... 3 more
17:29:09 Caused by: net.rubygrapefruit.platform.NativeException: Could not start '/usr/lib/jvm/zulu-8-amd64/bin/java'
17:29:09 	at net.rubygrapefruit.platform.internal.DefaultProcessLauncher.start(DefaultProcessLauncher.java:27)
17:29:09 	at net.rubygrapefruit.platform.internal.WrapperProcessLauncher.start(WrapperProcessLauncher.java:36)
17:29:09 	at org.gradle.process.internal.ExecHandleRunner.startProcess(ExecHandleRunner.java:98)
17:29:09 	at org.gradle.process.internal.ExecHandleRunner.run(ExecHandleRunner.java:71)
17:29:09 	... 4 more
17:29:09 Caused by: java.io.IOException: Cannot run program "/usr/lib/jvm/zulu-8-amd64/bin/java" (in directory "/agent/workspace/myproject"): error=7, Argument list too long
17:29:09 	at net.rubygrapefruit.platform.internal.DefaultProcessLauncher.start(DefaultProcessLauncher.java:25)
17:29:09 	... 7 more
17:29:09 Caused by: java.io.IOException: error=7, Argument list too long
```

While path jar is supported in gradle and can be overwritten via `      System.setProperty("org.gradle.internal.cmdline.max.length", CMDLINE_MAX_LENGTH);`, it is not enough since the classpath for the jar is small as it contains only the spotbugs/findbugs jars

 https://github.com/gradle/gradle/blob/master/subprojects/core/src/main/java/org/gradle/process/internal/util/LongCommandLineDetectionUtil.java#L45

which is invoked by https://github.com/gradle/gradle/blob/9e347dcbd00483d796e83c05c8b48d61dad06d0c/subprojects/core/src/main/java/org/gradle/process/internal/JavaExecHandleBuilder.java#L388 to determine if it should do path jar or not

However, the problem comes with java processes trying to use a long `auxclasspath`.

This is a sample of simple project:

```
0 = "-sortByClass"
1 = "-timestampNow"
2 = "-auxclasspath"
3 = "/Users/rperezalcolea/.gradle/caches/modules-2/files-2.1/com.google.guava/guava/19.0/6ce200f6b23222af3d8abb6b6459e6c44f4bb0e9/guava-19.0.jar"
4 = "-sourcepath"
5 = "/Users/rperezalcolea/Projects/bb/nebula/nebula/subprojects/core/build/nebulatest/netflix.nebula.spotbugs.NebulaSpotbugsCiPluginLauncherSpec/should-produce-html-reports-by-default/src/main/resources:/Users/rperezalcolea/Projects/bb/nebula/nebula/subprojects/core/build/nebulatest/netflix.nebula.spotbugs.NebulaSpotbugsCiPluginLauncherSpec/should-produce-html-reports-by-default/src/main/java"
6 = "-html"
7 = "-outputFile"
8 = "/Users/rperezalcolea/Projects/bb/nebula/nebula/subprojects/core/build/nebulatest/netflix.nebula.spotbugs.NebulaSpotbugsCiPluginLauncherSpec/should-produce-html-reports-by-default/build/reports/spotbugs/main.html"
9 = "-projectName"
10 = "should-produce-html-reports-by-default (spotbugsMain)"
11 = "-release"
12 = "unspecified"
13 = "-analyzeFromFile"
14 = "/var/folders/hb/c0ghc68d7vn26h4ny84kzv0r0000gn/T/spotbugs-gradle-plugin3618879618624022328.txt"
```

When these arguments are longer than `ARG_MAX` (2097152 on linux / 262144 OSX / 32767 windows) the process will fail.

### DIFF
--- a/src/main/groovy/com/github/spotbugs/snom/internal/SpotBugsRunner.java
+++ b/src/main/groovy/com/github/spotbugs/snom/internal/SpotBugsRunner.java
@@ -17,6 +17,7 @@ import com.github.spotbugs.snom.SpotBugsReport;
 import com.github.spotbugs.snom.SpotBugsTask;
 import edu.umd.cs.findbugs.annotations.NonNull;
 import java.io.File;
+import java.io.FileWriter;
 import java.io.IOException;
 import java.nio.charset.StandardCharsets;
 import java.nio.file.Files;
@@ -49,8 +50,15 @@ public abstract class SpotBugsRunner {
     args.add("-sortByClass");
     args.add("-timestampNow");
     if (!task.getAuxClassPaths().isEmpty()) {
-      args.add("-auxclasspath");
-      args.add(join(task.getAuxClassPaths().getFiles()));
+      if (task.getUseAuxclasspathFile().get()) {
+        args.add("-auxclasspathFromFile");
+        String auxClasspathFile = createFileForAuxClasspath(task.getAuxClassPaths().getFiles());
+        log.debug("Using auxclasspath file: {}", auxClasspathFile);
+        args.add(auxClasspathFile);
+      } else {
+        args.add("-auxclasspath");
+        args.add(join(task.getAuxClassPaths().getFiles()));
+      }
     }
     if (!task.getSourceDirs().isEmpty()) {
       args.add("-sourcepath");
@@ -106,6 +114,22 @@ public abstract class SpotBugsRunner {
     args.addAll(task.getExtraArgs().getOrElse(Collections.emptyList()));
     log.debug("Arguments for SpotBugs are generated: {}", args);
     return args;
+  }
+
+  private String createFileForAuxClasspath(Set<File> auxClasspathFiles) {
+    String auxClasspath =
+        auxClasspathFiles.stream().map(File::getAbsolutePath).collect(Collectors.joining("\n"));
+    try {
+      File tempFile = File.createTempFile("temp", "-spotbugs-auxclasspath");
+      FileWriter tempFileWriter = new FileWriter(tempFile);
+      tempFileWriter.write(auxClasspath);
+      tempFileWriter.close();
+      tempFile.deleteOnExit();
+      return tempFile.getAbsolutePath();
+    } catch (Exception e) {
+      // oops
+      throw new GradleException("Could not create auxiliary classpath file for SpotBugsTask");
+    }
   }
 
   private File generateFile(FileCollection files) {
--- a/src/functionalTest/groovy/com/github/spotbugs/snom/StandardFunctionalTest.groovy
+++ b/src/functionalTest/groovy/com/github/spotbugs/snom/StandardFunctionalTest.groovy
@@ -451,4 +451,45 @@ public class FooTest {
         result.output.contains("Applying com.h3xstream.findsecbugs.PredictableRandomDetector to FooTest")
         !result.output.contains("Trying to add already registered factory")
     }
+
+    def "can apply plugin using useAuxclasspathFile flag"() {
+        given:
+        buildFile << """
+spotbugs {
+  useAuxclasspathFile = true
+}
+dependencies {
+  implementation 'com.google.guava:guava:19.0'
+}"""
+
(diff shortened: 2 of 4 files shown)


rpalcolea:
Hi @KengoTODA do you have any thoughts/feedback on this?

src/main/groovy/com/github/spotbugs/snom/internal/SpotBugsRunner.java
```diff
@@ -108,6 +116,22 @@
...
+    try {
+      File tempFile = File.createTempFile("temp", "-spotbugs-auxclasspath");
+      FileWriter tempFileWriter = new FileWriter(tempFile);
+      tempFileWriter.write(auxClasspath);
+      tempFileWriter.close();
+      tempFile.deleteOnExit();
+      return tempFile.getAbsolutePath();
+    } catch (Exception e) {
+      // oops
+      throw new GradleException("Could not create auxiliary classpath file for SpotBugsTask");
```
wreulicke, line 131: The original stack trace is lost.
You should pass as the second argument.

```suggestion
      throw new GradleException("Could not create auxiliary classpath file for SpotBugsTask", e);
```

rpalcolea: Thanks for the suggestion @wreulicke 

Should be ready again

rpalcolea:
Hi @KengoTODA , I fix the merging conflicts for this one 👍

src/main/groovy/com/github/spotbugs/snom/internal/SpotBugsRunner.java
```diff
@@ -108,6 +116,22 @@
     return args;
   }
 
+  private String createFileForAuxClasspath(Set<File> auxClasspathFiles) {
+    String auxClasspath =
+        auxClasspathFiles.stream().map(File::getAbsolutePath).collect(Collectors.joining("\n"));
+    try {
+      File tempFile = File.createTempFile("temp", "-spotbugs-auxclasspath");
```
wreulicke, line 123: I concern a little that it leads to fill in temp files in a local disk.
But, it can not fill soon (the file might be small).
It will be great If it can be persisted project-locally.

@KengoTODA 
What do you think?

rpalcolea: Hi @wreulicke , good point.

I changed this to write to the `build` folder instead: https://github.com/spotbugs/spotbugs-gradle-plugin/commit/2b9f18cf2adf2534ceeaff1e8e5df775cf8ba9a2

wreulicke:
@KengoTODA 
ah, I have no permission to merge. 
Could you merge this PR?

KengoTODA:
Thanks for your PR! It'll be released soon.

Issue #245, opened 2020-05-11
Spotbugs report path should be provided in output when process fails


rpalcolea:
Hi folks,

Currently when running spotbugs and getting a failure, you can see the following:

```
3:38:34 * What went wrong:
03:38:34 Execution failed for task ':server:spotbugsMain'.
03:38:34 > A failure occurred while executing com.github.spotbugs.snom.internal.SpotBugsRunnerForWorker$SpotBugsExecutor
03:38:34    > Verification failed: SpotBugs execution thrown exception
03:38:34 
``` 

In order to get more details, `--stacktrace` has to be enabled and also you don't get details on where the report was added, if it was XML or HTML or something else.

I think it would be nice if the plugin outputs the full path when the execution fails

KengoTODA:
released in v4.1.0

PR #246, opened 2020-05-11
feat: show spotbugs report full path when spotbugs execution fails

rpalcolea:
Hi folks, this is for https://github.com/spotbugs/spotbugs-gradle-plugin/issues/245 

"Spotbugs report path should be provided in output when process fails"

Currently when running spotbugs and getting a failure, you can see the following:

```
3:38:34 * What went wrong:
03:38:34 Execution failed for task ':server:spotbugsMain'.
03:38:34 > A failure occurred while executing com.github.spotbugs.snom.internal.SpotBugsRunnerForWorker$SpotBugsExecutor
03:38:34    > Verification failed: SpotBugs execution thrown exception
03:38:34 
``` 

In order to get more details, `--stacktrace` has to be enabled and also you don't get details on where the report was added, if it was XML or HTML or something else.

I think it would be nice if the plugin outputs the full path when the execution fails

### DIFF
--- a/src/main/groovy/com/github/spotbugs/snom/internal/SpotBugsRunnerForWorker.java
+++ b/src/main/groovy/com/github/spotbugs/snom/internal/SpotBugsRunnerForWorker.java
@@ -104,11 +104,11 @@ public class SpotBugsRunnerForWorker extends SpotBugsRunner {
                     + findReportPath());
           }
         }
-      } catch (GradleException gradleException) {
+      } catch (GradleException e) {
         if (params.getIgnoreFailures().getOrElse(Boolean.FALSE).booleanValue()) {
-          log.warn("SpotBugs reported failures", gradleException);
+          log.warn("SpotBugs reported failures", e);
         } else {
-          throw gradleException;
+          throw e;
         }
       } catch (Exception e) {
         throw new GradleException("Verification failed: SpotBugs execution thrown exception", e);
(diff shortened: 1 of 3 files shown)


rpalcolea:
Hi @KengoTODA do you have any thoughts/feedback on this?

src/main/groovy/com/github/spotbugs/snom/internal/SpotBugsRunnerForWorker.java
```diff
@@ -91,18 +92,36 @@ public void execute() {
...
             throw new GradleException(
-                "Verification failed: SpotBugs violation found: " + findBugs2.getBugCount());
+                "Verification failed: SpotBugs violation found: "
+                    + findBugs2.getBugCount()
+                    + ". SpotBugs report can be found in "
+                    + findReportPath());
           }
         }
-      } catch (Exception e) {
+      } catch (GradleException gradleException) {
```
wreulicke, line 107: IMO, Use just e as the variable name for GradleException.
Do you have any concerns?

```suggestion
      } catch (GradleException gradleException) {
```

rpalcolea: Thanks for the suggestion @wreulicke 

Should be ready again

PR #250, opened 2020-05-14
Minor changes to simplify CI

KengoTODA:
1. Stop depending on auto-merge, because dependabot also provides a similar function
2. Remove Ghooks which isn't working as expected (after ghooks runs, husky will not run because `.git` directory already has hook script files) 
3. Bump up the version of Gradle for testing


KengoTODA:
@spotbugs/everyone please review this change to merge. Thanks in advance!

Issue #90, opened 2018-12-24
Unable to use Spotbugs in Android application


ZOlbrys:
Hey all,

I've recently tried to get Spotbugs working in a sample Android application and I haven't had any luck.

Here's a sample project showing my attempt: https://github.com/ZOlbrys/SpotbugsExample

When I try to run Spotbugs via gradle, I see this error:

```
Zachs-MBP:SpotbugsExample Zach$ ./gradlew spotbugs --stacktrace

FAILURE: Build failed with an exception.

* What went wrong:
Error while evaluating property 'allSource' of task ':app:spotbugs'
> path may not be null or empty string. path='null'

* Try:
Run with --info or --debug option to get more log output. Run with --scan to get full insights.

* Exception is:
org.gradle.api.InvalidUserDataException: Error while evaluating property 'allSource' of task ':app:spotbugs'
        at org.gradle.api.internal.tasks.properties.GetInputPropertiesVisitor$1.create(GetInputPropertiesVisitor.java:58)
        at org.gradle.api.internal.tasks.properties.GetInputPropertiesVisitor$1.create(GetInputPropertiesVisitor.java:48)
        at org.gradle.api.internal.changedetection.state.CacheBackedTaskHistoryRepository.snapshotTaskInputProperties(CacheBackedTaskHistoryRepository.java:303)
        at org.gradle.api.internal.changedetection.state.CacheBackedTaskHistoryRepository.createExecution(CacheBackedTaskHistoryRepository.java:146)
        at org.gradle.api.internal.changedetection.state.CacheBackedTaskHistoryRepository.access$100(CacheBackedTaskHistoryRepository.java:61)
        at org.gradle.api.internal.changedetection.state.CacheBackedTaskHistoryRepository$1.getCurrentExecution(CacheBackedTaskHistoryRepository.java:111)
        at org.gradle.api.internal.changedetection.changes.DefaultTaskArtifactStateRepository$TaskArtifactStateImpl.getStates(DefaultTaskArtifactStateRepository.java:208)
        at org.gradle.api.internal.changedetection.changes.DefaultTaskArtifactStateRepository$TaskArtifactStateImpl.isUpToDate(DefaultTaskArtifactStateRepository.java:93)
        at org.gradle.api.internal.tasks.execution.SkipUpToDateTaskExecuter.execute(SkipUpToDateTaskExecuter.java:50)
[...]

PR #252, opened 2020-05-14
Support Android Gradle Plugin 3.5.3, 3.6.3, 4.0.0

davidburstromspotify:
Fixes #233 and #90


davidburstromspotify:
@KengoTODA My suggestion regarding the Gradle versions is to bump 5.6 to 5.6.4 for CI, as any project that's still on 5.6 should be updating to 5.6.4 anyway. If that's not agreeable, the functional test would have to be modified to be skipped if Gradle 5.6 is used.

src/main/groovy/com/github/spotbugs/snom/internal/SpotBugsTaskFactory.java
```diff
@@ -63,30 +68,35 @@ private void generateForJava(Project project, Action<? super SpotBugsTask> confi
...
-                                  SpotBugsTask.class,
-                                  spotbugsTask -> {
-                                    spotbugsTask.setSourceDirs(task.getSource());
-                                    spotbugsTask.setClassDirs(
-                                        task.getOutputDirectory().getAsFileTree());
-                                    spotbugsTask.setAuxClassPaths(task.getClasspath());
-                                    configurationAction.execute(spotbugsTask);
-                                  });
-                        }));
+    if (project.getPlugins().hasPlugin("com.android.base")) {
```
KengoTODA, line 71: better to use [the `withId()` method](https://docs.gradle.org/5.6/javadoc/org/gradle/api/plugins/PluginContainer.html#withId-java.lang.String-org.gradle.api.Action-), then we do not need to care the order of loading plugin.

src/main/groovy/com/github/spotbugs/snom/internal/SpotBugsTaskFactory.java
```diff
@@ -63,30 +68,35 @@ private void generateForJava(Project project, Action<? super SpotBugsTask> confi
...
+
+      final BaseExtension baseExtension = project.getExtensions().getByType(BaseExtension.class);
+      DomainObjectSet<? extends BaseVariant> variants;
+      if (baseExtension instanceof AppExtension) {
+        variants = ((AppExtension) baseExtension).getApplicationVariants();
+      } else if (baseExtension instanceof LibraryExtension) {
+        variants = ((LibraryExtension) baseExtension).getLibraryVariants();
+      } else {
+        throw new RuntimeException("Unrecognized Android extension " + baseExtension);
```
KengoTODA, line 80: Better to throw an instance of [`GradleException`](https://docs.gradle.org/current/javadoc/org/gradle/api/GradleException.html) or its subclasses, but I cannot find suitable class at this moment.

otiai10:
Hi @davidburstromspotify, are you gonna work on this?

davidburstromspotify:
遅くなってごめんね。I started doing the changes, plus ensuring that AGP 4.0.0 is also supported.

src/main/groovy/com/github/spotbugs/snom/internal/SpotBugsTaskFactory.java
davidburstromspotify: Although I would agree with you generally, it unfortunately is not that simple for the AGP. If you use `withId()` on the base plugin, it will try to evaluate the block before the actual application/library plugin has applied.

That being said, I agree that it's better to find a solution where the loading order is not relevant, so then I need to look for more specific AGP plugins.

src/main/groovy/com/github/spotbugs/snom/internal/SpotBugsTaskFactory.java
davidburstromspotify: Agreed

Issue #253, opened 2020-05-14
Support for URL in the filter rules file just like maven FindBugs plugin


msiemczyk:
The company I work for is moving build system from maven to Gradle and we are trying to move all our functionality over. In maven we are still using findbugs-maven-plugin, but we load our include rules from SonarQube server that is also used with SonarLint.

![image](https://user-images.githubusercontent.com/5693250/81995254-c9117480-9617-11ea-9cd9-38bf713592d6.png)

We were planning to eventually upgrade to SpotBugs, but we never got around to it, and it also looks like SpotBugs maven plugin does not support URL.

https://github.com/spotbugs/spotbugs/issues/912

I'm not exactly sure if this support should be added to the core SpotBugs library or to each individual plugins, but I have decided to just add it to the Gradle plugin. I've tested it quite a bit added couple functional tests to cover the new use cases. Mind you I'm not using HTTP URL as that would involve at least some sort of mock web server, but simply using "file://" URL for tests. From automated testing point of view it really does not matter though.

Here is how you can configure it in gradle.build that would match our maven setup:

![image](https://user-images.githubusercontent.com/5693250/81995475-78e6e200-9618-11ea-8f63-936414b1900f.png)

I hope I didn't totally mess it up for you guys, but idea was to use Property<Object> so that the either URL or File can be given. If the URL is given then the rules file is downloaded to the temp location first and plugin uses it from there.

msiemczyk:
PR: https://github.com/spotbugs/spotbugs-gradle-plugin/pull/254

PR #254, opened 2020-05-14
feat: Support for URL in the filter rules file just like maven FindBugs plugin #253

msiemczyk:
Use Property so that the either URL or File can be given. If the URL is given then the rules file is downloaded to the temp location first and plugin uses it from there. Also added two functional tests to cover the new use cases.

### DIFF
--- a/src/main/groovy/com/github/spotbugs/snom/SpotBugsExtension.groovy
+++ b/src/main/groovy/com/github/spotbugs/snom/SpotBugsExtension.groovy
@@ -89,7 +89,7 @@ class SpotBugsExtension {
      */
     @NonNull
     final DirectoryProperty reportsDir;
-    
+
     /**
      * Property to set the filter file to limit which bug should be reported.
      *
@@ -100,7 +100,7 @@ class SpotBugsExtension {
      */
     @NonNull
     final Property<Object> includeFilter;
-    
+
     /**
      * Property to set the filter file to limit which bug should be reported.
      *
@@ -111,7 +111,7 @@ class SpotBugsExtension {
      */
     @NonNull
     final Property<Object> excludeFilter;
-    
+
     /**
      * Property to specify the target classes for analysis. Default value is empty that means all classes are analyzed.
      */
--- a/src/main/groovy/com/github/spotbugs/snom/SpotBugsExtension.groovy
+++ b/src/main/groovy/com/github/spotbugs/snom/SpotBugsExtension.groovy
@@ -40,8 +40,8 @@ import org.gradle.api.provider.Property;
  * &nbsp;&nbsp;&nbsp;&nbsp;visitors = [ 'FindSqlInjection', 'SwitchFallthrough' ]<br>
  * &nbsp;&nbsp;&nbsp;&nbsp;omitVisitors = [ 'FindNonShortCircuit' ]<br>
  * &nbsp;&nbsp;&nbsp;&nbsp;reportsDir = file("$buildDir/reports/spotbugs")<br>
- * &nbsp;&nbsp;&nbsp;&nbsp;includeFilter = file('spotbugs-include.xml')<br>
- * &nbsp;&nbsp;&nbsp;&nbsp;excludeFilter = file('spotbugs-exclude.xml')<br>
+ * &nbsp;&nbsp;&nbsp;&nbsp;includeFilter = file('spotbugs-include.xml') or new URL('http://sonar?findbugs-include.xml')<br>
+ * &nbsp;&nbsp;&nbsp;&nbsp;excludeFilter = file('spotbugs-exclude.xml') or new URL('http://sonar?findbugs-exclude.xml')<br>
  * &nbsp;&nbsp;&nbsp;&nbsp;onlyAnalyze = ['com.foobar.MyClass', 'com.foobar.mypkg.*']<br>
  * &nbsp;&nbsp;&nbsp;&nbsp;projectName = name<br>
  * &nbsp;&nbsp;&nbsp;&nbsp;release = version<br>
@@ -89,6 +89,7 @@ class SpotBugsExtension {
      */
     @NonNull
     final DirectoryProperty reportsDir;
+    
     /**
      * Property to set the filter file to limit which bug should be reported.
      *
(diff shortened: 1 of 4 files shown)


src/main/groovy/com/github/spotbugs/snom/SpotBugsExtension.groovy
```diff
@@ -40,8 +40,8 @@ import org.gradle.api.provider.Property;
  * &nbsp;&nbsp;&nbsp;&nbsp;visitors = [ 'FindSqlInjection', 'SwitchFallthrough' ]<br>
  * &nbsp;&nbsp;&nbsp;&nbsp;omitVisitors = [ 'FindNonShortCircuit' ]<br>
  * &nbsp;&nbsp;&nbsp;&nbsp;reportsDir = file("$buildDir/reports/spotbugs")<br>
- * &nbsp;&nbsp;&nbsp;&nbsp;includeFilter = file('spotbugs-include.xml')<br>
- * &nbsp;&nbsp;&nbsp;&nbsp;excludeFilter = file('spotbugs-exclude.xml')<br>
+ * &nbsp;&nbsp;&nbsp;&nbsp;includeFilter = file('spotbugs-include.xml') or new URL('http://sonar?findbugs-include.xml')<br>
+ * &nbsp;&nbsp;&nbsp;&nbsp;excludeFilter = file('spotbugs-exclude.xml') or new URL('http://sonar?findbugs-exclude.xml')<br>
```
KengoTODA, line 44: better to use valid URL like `https://example.com/path/to/findbugs-include.xml`

src/main/groovy/com/github/spotbugs/snom/SpotBugsExtension.groovy
```diff
@@ -40,8 +40,8 @@ import org.gradle.api.provider.Property;
  * &nbsp;&nbsp;&nbsp;&nbsp;visitors = [ 'FindSqlInjection', 'SwitchFallthrough' ]<br>
  * &nbsp;&nbsp;&nbsp;&nbsp;omitVisitors = [ 'FindNonShortCircuit' ]<br>
  * &nbsp;&nbsp;&nbsp;&nbsp;reportsDir = file("$buildDir/reports/spotbugs")<br>
- * &nbsp;&nbsp;&nbsp;&nbsp;includeFilter = file('spotbugs-include.xml')<br>
- * &nbsp;&nbsp;&nbsp;&nbsp;excludeFilter = file('spotbugs-exclude.xml')<br>
+ * &nbsp;&nbsp;&nbsp;&nbsp;includeFilter = file('spotbugs-include.xml') or new URL('http://sonar?findbugs-include.xml')<br>
```
KengoTODA, line 43: could you use comments to explain the usage of URL, or this snippet becomes invalid groovy script so user cannot copy-and-paste.

src/main/groovy/com/github/spotbugs/snom/SpotBugsExtension.groovy
```diff
@@ -98,7 +99,8 @@ class SpotBugsExtension {
      * <p>See also <a href="https://spotbugs.readthedocs.io/en/stable/filter.html">SpotBugs Manual about Filter file</a>.</p>
      */
     @NonNull
-    final RegularFileProperty includeFilter;
+    final Property<Object> includeFilter;
```
KengoTODA, line 102: this field is public API, so it's better to keep it compatible. I suggest you to add a new field, and link `includeFilter` prop to it via a convention. like:

```groovy
includeFilter.convention(() -> downloadFrom(includeFilterUrl))
```

or, it's better to break backward compatibility, make next release as a major update (v5.0.0). Refer https://www.conventionalcommits.org/en/v1.0.0/ to learn how to release `BREAKING CHANGE`.

src/main/groovy/com/github/spotbugs/snom/SpotBugsTask.groovy
```diff
@@ -416,4 +429,28 @@ class SpotBugsTask extends DefaultTask implements VerificationTask {
...
+                tempFile.deleteOnExit()
+
+                tempFile.withOutputStream { out ->
+                    out << filter.openStream()
+                }
+
+                filterFileProperty.set(tempFile)
+            }
+            else {
+                throw new IllegalArgumentException("The '$filterName' property must be a File or URL.")
```
KengoTODA, line 452: This exception is thrown to Gradle core (our code does not catch this exception), then I suggest using a subclass of `GradleException` like [InvalidUserDataException](https://docs.gradle.org/current/javadoc/org/gradle/api/InvalidUserDataException.html).

msiemczyk:
@KengoTODA Thank you for your feedback, but I already spent too much time on this and I don't really have more time to make the changes you requested. Also we wouldn't be able to wait to version 5.0 in order to use this functionality; we are moving our build system now and I need to restore all out features from maven. For these reasons I've already built and published custom version of this plugin to our own maven and will just use that until the public version of this plugins supports URL for filter files.

Of course you are always welcome to make these changes yourself to enhance the plugin for others.

src/main/groovy/com/github/spotbugs/snom/SpotBugsTask.groovy
msiemczyk: What difference will that make?

src/main/groovy/com/github/spotbugs/snom/SpotBugsExtension.groovy
msiemczyk: I don't understand your comment by keeping it compatible. The property is backwards compatible the way I wrote the code. All you loose is strongly typed RegularFileProperty, but I didn't see any other way for this one setting to accept to different types.

If you add new property, includeFilterUrl, then you will also need extra user input validation to make sure both things are not specified at the same time. Also I don't really understand that "convention()" thing and Gradle has rather poor documentation.

src/main/groovy/com/github/spotbugs/snom/SpotBugsExtension.groovy
msiemczyk: So you would want the '...or new URL' part removed from here and added below instead?

msiemczyk: So you would want the '...or new URL' part removed from here and added below instead?

PR #256, opened 2020-05-18
Stop printing null if no report is available

KengoTODA:
When user uses the `base` plugin, it's possible to have no reporting so `findReportPath()` returns `null`. This PR will fix this behavior, by simply skipping the report path.

### DIFF
--- a/src/functionalTest/groovy/com/github/spotbugs/snom/BasePluginFunctionalTest.groovy
+++ b/src/functionalTest/groovy/com/github/spotbugs/snom/BasePluginFunctionalTest.groovy
@@ -141,7 +141,7 @@ task spotbugsMain(type: com.github.spotbugs.snom.SpotBugsTask) {
 """
         when:
         def arguments = [':spotbugsMain']
-        if(!isWorkerApi) {
+        if (!isWorkerApi) {
             arguments.add('-Pcom.github.spotbugs.snom.worker=false')
         }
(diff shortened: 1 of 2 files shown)


KengoTODA:
@spotbugs/everyone please review this change to merge. Thanks in advance!

src/functionalTest/groovy/com/github/spotbugs/snom/BasePluginFunctionalTest.groovy
```diff
@@ -123,4 +123,42 @@ task spotbugsMain(type: com.github.spotbugs.snom.SpotBugsTask) {
...
+        buildFile << """
+task spotbugsMain(type: com.github.spotbugs.snom.SpotBugsTask) {
+    dependsOn 'classes'
+    classDirs = sourceSets.main.output
+}
+"""
+        when:
+        def arguments = [':spotbugsMain']
+        if(!isWorkerApi) {
+            arguments.add('-Pcom.github.spotbugs.snom.worker=false')
```
jscancella, line 145: maybe add a comment on why we need to turn off the worker api? Just looking at the code it isn't obvious to me.

KengoTODA: added a comment, please check 42802b6

PR #258, opened 2020-05-20
fix: SpotBugsTask was not cacheable

### LINKED ISSUE #244: Enable build cache support
Plugin version 4.0 lost [gradle build cache](https://docs.gradle.org/current/userguide/build_cache.html) support.

It was enabled in 3.0.0: https://github.com/spotbugs/spotbugs-gradle-plugin/blob/3.0.0/src/main/java/com/github/spotbugs/SpotBugsTask.java#L59

The fix is just to get back the `@CacheableTask` annotation on the `SpotBugsTask`.

Build cache can be verified as:

1. Run spotbugs task with build cache enabled: 
`gradlew spotbugsMain --build-cache`
task status will be SUCCESS (task executed)
2. Do cleanup
`gradlew clean`
3. Run the same task with enabled cache:
`gradlew spotbugsMain --build-cache`
task status should be FROM_CACHE (task output taken from the cache without execution)

Buzz-Lightyear:
close #244

### DIFF
--- a/src/main/groovy/com/github/spotbugs/snom/SpotBugsTask.groovy
+++ b/src/main/groovy/com/github/spotbugs/snom/SpotBugsTask.groovy
@@ -35,6 +35,7 @@ import org.gradle.api.file.FileCollection;
 import org.gradle.api.model.ObjectFactory;
 import org.gradle.api.provider.ListProperty;
 import org.gradle.api.provider.Property;
+import org.gradle.api.tasks.CacheableTask
 import org.gradle.api.tasks.Input;
 import org.gradle.api.tasks.InputFile;
 import org.gradle.api.tasks.InputFiles;
@@ -82,6 +83,8 @@ import javax.inject.Inject
  *
  * <p>See also <a href="https://spotbugs.readthedocs.io/en/stable/running.html">SpotBugs Manual about configuration</a>.</p>
  */
+
+@CacheableTask
 class SpotBugsTask extends DefaultTask implements VerificationTask {
     private static final String FEATURE_FLAG_WORKER_API = "com.github.spotbugs.snom.worker";
     private final Logger log = LoggerFactory.getLogger(SpotBugsTask);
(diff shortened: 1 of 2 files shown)


src/main/groovy/com/github/spotbugs/snom/SpotBugsTask.groovy
```diff
@@ -175,7 +178,7 @@ class SpotBugsTask extends DefaultTask implements VerificationTask {
      * Property to specify the name of project. Some reporting formats use this property.
      * Default value is {@code "${project.name} (${task.name})"}.
      */
-    @Input
+    @Internal
```
KengoTODA, line 181: Why this change is necessary? It seems needless for me. 🤔

DPUkyle: By marking property `projectName` as an input, this has the ability to cause cache misses. I understand this value might show up in the output reports, but it has the ability to affect the cache. Marking as `@Internal` means it is not considered as part of the master cache key for tasks of type `SpotBugsTask`.

The test on this PR will definitely fail if we revert this line.

Another possibility is that we do not change this annotation.  However projects will potentially have a lowered cache rate - especially projects who do not explicitly set `rootProject.name = '...'` in their settings.gradle file. There should be documentation in this case saying that the default value for `projectName` can cause low cache hit %.

DPUkyle: Another possibility is the `@Console` annotation: https://docs.gradle.org/current/javadoc/org/gradle/api/tasks/Console.html

KengoTODA: Thanks! I understood.

I'm little bit confused because `projectName` is actually input of this task, and it affects not only console but also report file. So even though annotating with `@Internal` or `@Console` actually solves this problem, I feel little strange to use (not so semantic?).

I'll request author to add a comment to use `@Internal`, for future code reviewers.

PR #272, opened 2020-05-26
fix: create auxclasspath file for each spotbugs task

OdysseusLives:
and auxclasspath file is removed and created instead of truncated in place

This addresses #259 by avoiding re-creating the same file in any invocation, and adds some additional information if a failure occurs.

### DIFF
--- a/src/main/groovy/com/github/spotbugs/snom/internal/SpotBugsRunner.java
+++ b/src/main/groovy/com/github/spotbugs/snom/internal/SpotBugsRunner.java
@@ -128,13 +128,22 @@ public abstract class SpotBugsRunner {
               task.getProject().getBuildDir().getAbsolutePath(),
               "spotbugs",
               "spotbugs-auxclasspath-for-" + task.getName());
-      Files.createDirectories(auxClasspathFile.getParent());
-      Files.createFile(auxClasspathFile);
-      Files.write(auxClasspathFile, auxClasspath.getBytes(), StandardOpenOption.TRUNCATE_EXISTING);
-      return auxClasspathFile.normalize().toString();
+      try {
+        Files.createDirectories(auxClasspathFile.getParent());
+        if (Files.exists(auxClasspathFile)) {
+          Files.delete(auxClasspathFile);
+        }
+        Files.createFile(auxClasspathFile);
+        Files.write(auxClasspathFile, auxClasspath.getBytes(), StandardOpenOption.WRITE);
+        return auxClasspathFile.normalize().toString();
+      } catch (Exception e) {
+        throw new GradleException(
+            "Could not create auxiliary classpath file for SpotBugsTask at "
+                + auxClasspathFile.normalize().toString(),
+            e);
+      }
     } catch (Exception e) {
-      // oops
-      throw new GradleException("Could not create auxiliary classpath file for SpotBugsTask");
+      throw new GradleException("Could not create auxiliary classpath file for SpotBugsTask", e);
     }
   }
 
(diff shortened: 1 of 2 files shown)


src/main/groovy/com/github/spotbugs/snom/internal/SpotBugsRunner.java
```diff
@@ -127,14 +127,23 @@ private String createFileForAuxClasspath(SpotBugsTask task) {
           Paths.get(
               task.getProject().getBuildDir().getAbsolutePath(),
               "spotbugs",
-              "spotbugs-auxclasspath");
-      Files.createDirectories(auxClasspathFile.getParent());
-      Files.createFile(auxClasspathFile);
-      Files.write(auxClasspathFile, auxClasspath.getBytes(), StandardOpenOption.TRUNCATE_EXISTING);
-      return auxClasspathFile.normalize().toString();
+              "spotbugs-auxclasspath-for-" + task.getName());
```
KengoTODA, line 130: just feeling: `"auxclasspath", task.getName());` would be good replacement, then the directory hierarchy will be more intuitive like:

```
project/
  - build/
    - spotbugs/
      - auxclasspath/
        - spotbugsMain
        - spotbugsTest
```

src/main/groovy/com/github/spotbugs/snom/internal/SpotBugsRunner.java
```diff
@@ -127,14 +127,23 @@ private String createFileForAuxClasspath(SpotBugsTask task) {
...
+      try {
+        Files.createDirectories(auxClasspathFile.getParent());
+        if (Files.exists(auxClasspathFile)) {
+          Files.delete(auxClasspathFile);
+        }
+        Files.createFile(auxClasspathFile);
+        Files.write(auxClasspathFile, auxClasspath.getBytes(), StandardOpenOption.WRITE);
```
KengoTODA, line 137: What is our motivation to replace `TRUNCATE_EXISTING` with `WRITE`?
and I guess `TRUNCATE_EXISTING` lets us remove the lines 133-136? (sorry I didn't noticed it previously)

src/main/groovy/com/github/spotbugs/snom/internal/SpotBugsRunner.java
```diff
@@ -127,14 +127,23 @@ private String createFileForAuxClasspath(SpotBugsTask task) {
...
+      } catch (Exception e) {
+        throw new GradleException(
+            "Could not create auxiliary classpath file for SpotBugsTask at "
+                + auxClasspathFile.normalize().toString(),
+            e);
+      }
     } catch (Exception e) {
-      // oops
-      throw new GradleException("Could not create auxiliary classpath file for SpotBugsTask");
+      throw new GradleException("Could not create auxiliary classpath file for SpotBugsTask", e);
```
KengoTODA, line 146: 👍

src/main/groovy/com/github/spotbugs/snom/internal/SpotBugsRunner.java
OdysseusLives: The motivation was to delete and create the file each time, but the outcome should be taken care of with the directory hierarchy however.

src/main/groovy/com/github/spotbugs/snom/internal/SpotBugsRunner.java
OdysseusLives: I like that 👍

Issue #277, opened 2020-06-02
Reintroduce SpotBugsTask#classpath property for backwards-compatibility


DPUkyle:
In SpotBugsTask v4 we use a property called `auxClassPaths` to parameterize a similarly-named CLI argument to the spotbugs tool. The equivalent argument in plugin v3 and prior was `classpath`.

In my workplace, we use a binary plugin to configure the SpotBugsTask globally, similar to: 
```
tasks.withType(SpotBugsTask.class).each( t -> {
  t.setClasses(...);
  t.setClasspath(...);
});
```

So the `setClasspath` above must be changed to `setAuxClassPaths(...)`, but that breaks compatibility with those who use an older version of the plugin.

My suggestion is to re-add the `classpath` property as a deprecated property, additionally annotating it with [ReplacedBy](https://docs.gradle.org/current/javadoc/org/gradle/api/model/ReplacedBy.html).

The deprecated `classpath` will just act as a pass-through to the proper `auxClassPaths` accessors/mutators.

jscancella:
Instead of creating setters and getters that are deprecated and just point to the current property, can't you detect which version of the plugin they are using and use either `classpath` or `auxClassPath`?

DPUkyle:
Closing per #278

PR #278, opened 2020-06-02
fix: Reintroduce SpotBugsTask#classpath property (#277)

DPUkyle:
Provided backwards-compatibility for configuring old and new versions of the plugin.

Obviously, new spotbugs users should not use this property and we should remove it in the next major version.

### DIFF
--- a/src/main/groovy/com/github/spotbugs/snom/SpotBugsTask.groovy
+++ b/src/main/groovy/com/github/spotbugs/snom/SpotBugsTask.groovy
@@ -23,6 +23,7 @@ import edu.umd.cs.findbugs.annotations.Nullable;
 import edu.umd.cs.findbugs.annotations.OverrideMustInvoke
 import org.gradle.api.file.DirectoryProperty
 import org.gradle.api.file.RegularFileProperty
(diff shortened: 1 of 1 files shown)


DPUkyle:
Yup, all valid points.

As a workaround, I can do this:
* No more referring to the task by FQCN, `c.g.s.SBT` vs `c.g.s.snom.SBT` using [TaskContainer#withType(Class, Action)](https://docs.gradle.org/current/javadoc/org/gradle/api/DomainObjectCollection.html#withType-java.lang.Class-org.gradle.api.Action-)
* Have to use less-reliable `tasks.getByName` or `named`
* Use reflection to detect old vs. new, then call [Task#setProperty(String, Object](https://docs.gradle.org/current/javadoc/org/gradle/api/Task.html#setProperty-java.lang.String-java.lang.Object-) as appropriate. I'm not 100% sure `setProperty` does what I want.
* If `setProperty` fails, I can always fall back to traditional reflection.

The worst part of this is the loss of `tasks.withType...` as a convenience to set common properties for both the main and test tasks. Not a big deal.

I will close and reopen if there's a major problem with the above.

Issue #286, opened 2020-06-18
report location is URL


jscancella:
As per the request here https://github.com/spotbugs/spotbugs-gradle-plugin/issues/284#issuecomment-646267149

You want to target this line https://github.com/spotbugs/spotbugs-gradle-plugin/blob/master/src/main/groovy/com/github/spotbugs/snom/internal/SpotBugsRunnerForJavaExec.java#L44
instead do something like .map(File::toURI)

PR #287, opened 2020-06-19
fix: Print report location as file uri and improve warning messages

cmuchinsky:
* warning messages are now consistent with those reported by other native Gradle static code analysis tools (PMD, Checkstyle, etc...).
* fixes https://github.com/spotbugs/spotbugs-gradle-plugin/issues/286

### DIFF
--- a/src/main/groovy/com/github/spotbugs/snom/internal/SpotBugsRunnerForWorker.java
+++ b/src/main/groovy/com/github/spotbugs/snom/internal/SpotBugsRunnerForWorker.java
@@ -20,6 +20,7 @@ import edu.umd.cs.findbugs.FindBugs2;
 import edu.umd.cs.findbugs.TextUICommandLine;
 import edu.umd.cs.findbugs.annotations.CheckForNull;
 import edu.umd.cs.findbugs.annotations.NonNull;
+import java.nio.file.Paths;
 import java.util.List;
 import java.util.Objects;
 import org.gradle.api.Action;
@@ -96,15 +97,13 @@ public class SpotBugsRunnerForWorker extends SpotBugsRunner {
           findBugs2.execute();
           if (findBugs2.getErrorCount() > 0) {
             throw new GradleException(
-                "Verification failed: SpotBugs error found: "
-                    + findBugs2.getErrorCount()
-                    + ". "
+                findBugs2.getErrorCount()
+                    + " SpotBugs errors were found. "
                     + buildMessageAboutReport());
           } else if (findBugs2.getBugCount() > 0) {
             throw new GradleException(
-                "Verification failed: SpotBugs violation found: "
-                    + findBugs2.getBugCount()
-                    + ". "
+                findBugs2.getBugCount()
+                    + " SpotBugs violations were found. "
                     + buildMessageAboutReport());
           }
         }
@@ -112,7 +111,7 @@ public class SpotBugsRunnerForWorker extends SpotBugsRunner {
         if (params.getIgnoreFailures().getOrElse(Boolean.FALSE).booleanValue()) {
           final boolean showStackTraces =
               params.getShowStackTraces().getOrElse(Boolean.TRUE).booleanValue();
-          log.warn("SpotBugs reported failures", showStackTraces ? e : e.getMessage());
+          log.warn(e.getMessage(), showStackTraces ? e : null);
         } else {
           throw e;
         }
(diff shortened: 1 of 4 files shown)


cmuchinsky:
With these changes, the output with stack traces disabled looks like:
```
> Task :spotbugsMain
1 SpotBugs violations were found. See the report at: file:///Users/redacted/build/reports/spotbugs/main.html
```

src/main/groovy/com/github/spotbugs/snom/internal/SpotBugsRunnerForWorker.java
```diff
@@ -96,23 +97,21 @@ public void execute() {
...
+                    + " SpotBugs violations were found. "
                     + buildMessageAboutReport());
           }
         }
       } catch (GradleException e) {
         if (params.getIgnoreFailures().getOrElse(Boolean.FALSE).booleanValue()) {
           final boolean showStackTraces =
               params.getShowStackTraces().getOrElse(Boolean.TRUE).booleanValue();
-          log.warn("SpotBugs reported failures", showStackTraces ? e : e.getMessage());
+          log.warn(e.getMessage(), showStackTraces ? e : null);
```
jscancella, line 114: If you are going to the trouble of redoing the output message I would like to see this changed from a ternary operator to something cleaner. 

Something along the line of 
```
@Override
    public void execute() {
      SpotBugsWorkParameters params = getParameters();
      String[] args = params.getArguments().get().toArray(new String[0]);
      DetectorFactoryCollection.resetInstance(new DetectorFactoryCollection());

      try {
        edu.umd.cs.findbugs.Version.printVersion(false);
        try (FindBugs2 findBugs2 = new FindBugs2()) {
          TextUICommandLine commandLine = new TextUICommandLine();
          FindBugs.processCommandLine(commandLine, args, findBugs2);
          findBugs2.execute();
          
          StringBuilder message = new StringBuilder();
          if (findBugs2.getErrorCount() > 0) {message.append(findBugs2.getErrorCount()).append(" SpotBugs errors were found. ");} 
          if (findBugs2.getBugCount() > 0) {message.append(findBugs2.getBugCount()).append(" SpotBugs violations were found. ");}
          if(message.length() > 0)
          {
            buildMessageAboutReport(message);
            GradleException e = new GradleException(message.toString());
            
            if(params.getIgnoreFailures().getOrElse(Boolean.FALSE)) {
              log.warn(message.toString());
              if(params.getShowStackTraces().getOrElse(Boolean.TRUE)) {
                log.warn("", e);
              }
            }
            else {
              throw e; //fail the gradle build
            }
          } 
              
        }
      } catch (Exception e) {
        throw new GradleException("Verification failed: SpotBugs execution thrown exception", e);
      }
      
    }

    private void buildMessageAboutReport(StringBuilder message) {
      String reportPath = findReportPath();
      if (reportPath != null) {
        message.append("SpotBugs report can be found in ").append(reportPath);
      }
    }

    @CheckForNull
    private String findReportPath() {
      List<String> arguments = getParameters().getArguments().get();
      int outputFileParameterIndex = arguments.indexOf("-outputFile");
      if (outputFileParameterIndex > 0) {
        return arguments.get(outputFileParameterIndex + 1);
      }
      return null;
    }
```

src/main/groovy/com/github/spotbugs/snom/internal/SpotBugsRunnerForWorker.java
```diff
@@ -94,42 +95,42 @@ public void execute() {
...
+          }
+          if (message.length() > 0) {
+            String reportPath = findReportPath();
+            if (reportPath != null) {
+              message.append(" See the report at: ").append(Paths.get(reportPath).toUri());
+            }
+
+            GradleException e = new GradleException(message.toString());
+
+            if (params.getIgnoreFailures().getOrElse(Boolean.FALSE).booleanValue()) {
```
jscancella, line 117: Sorry, but shouldn't there be a logical not here?

cmuchinsky: Not sure I follow what you are asking for, if `ignoreFailures` is true then warning messages are logged, otherwise the exception is thrown.

jscancella: you're right. I was thinking "ignore failures, but wait we are actually doing stuff instead of ignoring it!". You are correct. Everything looks good to me!

KengoTODA:
v4.4.3 contains this fix.
https://github.com/spotbugs/spotbugs-gradle-plugin/releases/tag/4.4.3

PR #290, opened 2020-06-24
fix: bump up the default toolVersion to 4.0.6

KengoTODA:
4.0.5 has a critical issue around BCEL usage, so bump up the default `toolVersion` to the fixed version.


KengoTODA:
@spotbugs/everyone please review this change, thank you!

Issue #279, opened 2020-06-02
SpotBugsTask#auxClassPaths may not be annotated correctly


DPUkyle:
I think `SpotBugsTask#auxClassPaths` _should_ be annotated as the following and return a `ConfigurableFileCollection`, but I'd like to have community confirmation before implementing.

```
    /**
     * Property to specify the aux class paths that contains the libraries to refer during analysis.
     * Default value is the compile-scope dependencies of the target sourceSet.
     */
    @Classpath
    ConfigurableFileCollection auxClassPaths
```

By declaring the return type as a `ConfigurableFileCollection`, we can use the convenient `from(...)` [API](https://docs.gradle.org/current/javadoc/org/gradle/api/file/ConfigurableFileCollection.html#from-java.lang.Object...-). But maybe this is too much as it also requires a `CFC` to set the value. So I'm not sure changing the type is the right move, but I feel strongly that the `@Classpath` annotation is correct.

jscancella:
Why not give it a try and submit a pull a request?

alaa-nassef:
@KengoTODA can we please merge #314 and have a new release, because this is affecting us. My project has about 90 cacheable spotbugs tasks that get executed every time because of this

KengoTODA:
I'm waiting the reply in #314, since I lack knowledge to make a judge. Thanks for your waiting.

PR #314, opened 2020-07-22
fix: Use `Classpath` normalization for `auxClassPaths`

bigdaz:
Declaring `auxClassPaths` as a `Classpath` property allows Gradle to perform additional normalization,
so that functionally-identical jar files do not trigger a re-execution of the SpotBugsTask. 

A re-generated jar containing identical classes will have different entry timestamps for each class entry,
and as such is not bitwise identical even though it is functionally identical.
Without this fix, changing an input jar file will change the input cache key, leading to consistent misses from the local build cache. 

See https://docs.gradle.org/6.0/userguide/more_about_tasks.html#sec:task_input_using_classpath_annotations

Note that I have tested this change locally, but not added any specific test coverage for this use case.

### DIFF
--- a/src/main/groovy/com/github/spotbugs/snom/SpotBugsTask.groovy
+++ b/src/main/groovy/com/github/spotbugs/snom/SpotBugsTask.groovy
@@ -36,6 +36,7 @@ import org.gradle.api.model.ObjectFactory;
 import org.gradle.api.provider.ListProperty;
 import org.gradle.api.provider.Property;
 import org.gradle.api.tasks.CacheableTask
+import org.gradle.api.tasks.Classpath
 import org.gradle.api.tasks.Input;
 import org.gradle.api.tasks.InputFile;
 import org.gradle.api.tasks.InputFiles;
@@ -235,7 +236,7 @@ class SpotBugsTask extends DefaultTask implements VerificationTask {
      * Default value is the compile-scope dependencies of the target sourceSet.
      */
     @InputFiles
-    @PathSensitive(PathSensitivity.RELATIVE)
+    @Classpath
     FileCollection auxClassPaths;
 
     /**


KengoTODA:
It seems that #279 is similar to this fix. @DPUkyle could you review this change?
If you know something to consider, please share it with us. :)

DPUkyle:
Yes, this is great. Thanks @bigdaz for fixing it!

src/main/groovy/com/github/spotbugs/snom/SpotBugsTask.groovy
```diff
@@ -235,7 +236,7 @@ class SpotBugsTask extends DefaultTask implements VerificationTask {
      * Default value is the compile-scope dependencies of the target sourceSet.
      */
     @InputFiles
```
DPUkyle, line 238: nit: we can remove `@InputFiles` as it's replaced by `@Classpath`

bigdaz: That's correct, [unless there is a goal to remain compatible with Gradle < 3.2](https://github.com/gradle/gradle/blob/ee65597ecffcfde800aedda9f4c956b688b956a8/subprojects/core-api/src/main/java/org/gradle/api/tasks/Classpath.java#L38-L39).

foto-andreas:
Is it correct, that v4.5.0 has both annotations (@InputFiles AND @Classpath)? Currently I am searching for changes of auxClassPaths fingerprints in my project...

bigdaz:
> Is it correct, that v4.5.0 has both annotations (@InputFiles AND @classpath)? 

It's safe to remove the `@InputFiles` annotation, [unless there is a goal to remain compatible with Gradle < 3.2](https://github.com/gradle/gradle/blob/ee65597ecffcfde800aedda9f4c956b688b956a8/subprojects/core-api/src/main/java/org/gradle/api/tasks/Classpath.java#L38-L39). I'll leave that to the plugin maintainers to decide.

KengoTODA:
This plugin does not support Gradle &lt; 3.2, so we can remove `@InputFiles` in our case.

PR #316, opened 2020-07-30
Support SARIF format

KengoTODA:
Support the SARIF format feature which is introduced from SpotBugs 4.1.


KengoTODA:
Hi @spotbugs/everyone, please review this change when you have time. Thanks in advance!

PR #354, opened 2020-09-30
fix: remove needless annotation

KengoTODA:
we don't support Gradle < 3.2, so the InputFiles annotation is needless 
for fields which is already annotated with the Classpath annotation.

refs https://github.com/spotbugs/spotbugs-gradle-plugin/pull/314#issuecomment-700773441


jscancella (approved):
LGTM

Issue #387, opened 2020-11-20
MissingPropertyException when setting effort of SpotBugsExtension


daanschipper:
When optimizing the configuration time using the gradle-profiler and checking the flame graph output I noticed the `SpotBugsExtension.setEffort()` method throwing `MissingPropertyException`s. I'm using `project.getExtensions().getByType(SpotBugsExtension.class).setEffort("max")` to set the effort. 

It's this line [`Effort effort = name == null ? null : Effort.valueOf(name.toUpperCase())`](https://github.com/spotbugs/spotbugs-gradle-plugin/blob/fd232f92e1c1750291b732563a879d7629cb11a2/src/main/groovy/com/github/spotbugs/snom/SpotBugsExtension.groovy#L199) causing the exception, but I don't have enough Groovy knowledge to understand why it is throwing the `MissingPropertyException`.

Here you can see the time it takes to apply the Spotbugs plugin and configure basic stuff of the extension. One third of the time is spent with `SpotbugsExtension.setEffort()`.

<img width="1192" alt="Screenshot 2020-11-20 at 17 06 03" src="https://user-images.githubusercontent.com/2778477/99821752-b477e100-2b52-11eb-9af1-60acf39e5921.png">

Unfortunately I cannot share the full stacktrace, but this is the final part of the stacktrace:

```
invokeStaticMissingProperty:1021, MetaClassImpl (groovy.lang)
getProperty:1866, MetaClassImpl (groovy.lang)
getProperty:1842, MetaClassImpl (groovy.lang)
invokeStaticMethod:1474, MetaClassImpl (groovy.lang)
callStatic:62, StaticMetaClassSite (org.codehaus.groovy.runtime.callsite)
callStatic:216, AbstractCallSite (org.codehaus.groovy.runtime.callsite)
valueOf:-1, Effort (com.github.spotbugs.snom)
call:-1, Effort$valueOf (com.github.spotbugs.snom)
setEffort:192, SpotBugsExtension (com.github.spotbugs.snom)
```

Environment:
Gradle 6.7
Java openjdk 11.0.6 2020-01-14
Spotbugs gradle plugin 4.5.0

jscancella:
I would think this would cause your build to fail since it is throwing an exception. Looking at the stack trace it seems it is somehow trying to set the `effort` to `-1`?

daanschipper:
Some dynamic groovy calls are happening, the exception is silently caught in `MetaClassImpl`
```
try {
[...]

PR #392, opened 2020-11-28
fix: get enum values without valueOf() (#387)

daanschipper:
to avoid silent MissingPropertyException thrown by Groovy

### DIFF
--- a/src/main/groovy/com/github/spotbugs/snom/SpotBugsExtension.groovy
+++ b/src/main/groovy/com/github/spotbugs/snom/SpotBugsExtension.groovy
@@ -191,13 +191,11 @@ class SpotBugsExtension {
     }
 
     void setReportLevel(@Nullable String name) {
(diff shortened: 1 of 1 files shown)


jscancella (changes requested):
Thanks for taking the time to submit this pull request. Please include a comment as to why we are doing it this way instead of `valueOf()` so that in the future when the issue is fixed this can be reverted or refactored. Please also include the URL of upstream the groovy/gradle ticket.

KengoTODA (changes requested):
This implementation still depends on a static method `values()`, then its invocation still throws `MissingPropertyException` based on [my research](https://github.com/spotbugs/spotbugs-gradle-plugin/issues/387#issuecomment-732081622). So I'm afraid that this change won't solve our problem. 🤔

daanschipper:
Thanks for double checking, I thought the issue was avoided by doing it this way. I'll close this pr then as it does not resolve the issue.

PR #419, opened 2021-02-08
Try the Gradle 7.0-milestone-2

KengoTODA:
[Gradle 7.0-milestone-1 has been published](https://docs.gradle.org/7.0-milestone-1/release-notes.html). In this branch, we'll confirm that it works with our plugin.

In my local, it reports an warning like below. We need to make the `:publishPluginJar` task depending on the `:processVersionFile` task explicitly.

```
> Task :publishPluginJar
Validation failed for task ':publishPluginJar', disabling optimizations:
  - Task ':publishPluginJar' uses the output of task ':processVersionFile', without declaring an explicit dependency (using Task.dependsOn() or Task.mustRunAfter()) or an implicit dependency (declaring task ':processVersionFile' as an input). This can lead to incorrect results being produced, depending on what order the tasks are executed.
```


KengoTODA:
#480 will resolve this issue

PR #425, opened 2021-02-11
Build optimization

KengoTODA:
Optimize Gradle build script for performance.

I also found that Groovydoc reports error because it fails parsing lambda in `.java` files.
We can solve it by migrating `.java` to `.groovy`, I will handle it in another PR.

### DIFF
--- a/build.gradle
+++ b/build.gradle
@@ -9,8 +9,6 @@ plugins {
     id 'org.sonarqube' version '3.1.1'
 }
 
-sourceCompatibility = 1.8
-targetCompatibility = 1.8
 group = 'com.github.spotbugs.snom'
 
 repositories {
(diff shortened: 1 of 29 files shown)


jscancella (approved):
LGTM

jscancella (approved):
Looks good to me. Though I find it interesting that using
```
tasks.named('processResources').configure {
    dependsOn processVersionFile
}
```
is faster than 
```
 tasks.processResources.dependsOn processVersionFile
```

How much did it speed it up? How much did changing gradle versions speed it up?

KengoTODA:
Thanks for reviewing this PR! :)

> Looks good to me. Though I find it interesting that using
> 
> ```
> tasks.named('processResources').configure {
>     dependsOn processVersionFile
> }
> ```
> 
> is faster than
> 
> ```
>  tasks.processResources.dependsOn processVersionFile
> ```
> 
> How much did it speed it up? How much did changing gradle versions speed it up?

`tasks.processResources` creates the `processResources` task even when it is needless (e.g. `./gradlew spotlessApply` does not need `processResources` task). `tasks.named('processResources').configure { ... }` will run when the target task is created by other tasks.

For detail, please check the official documentation about this topic: https://docs.gradle.org/current/userguide/task_configuration_avoidance.html

PR #429, opened 2021-02-16
feat: run java process from no-isolated workers

### LINKED ISSUE #416: Incorrect keepAliveMode used for Gradle worker daemon
Right now spotbug tasks are executed in a separate java fork and that fork is started with keepAliveMode=DAEMON option.

`Started Gradle worker daemon (0.999 secs) with fork options DaemonForkOptions{executable=/usr/lib/jvm/java-11-openjdk-11.0.6.10-1.el7_7.x86_64/bin/java, minHeapSize=null, maxHeapSize=null, jvmArgs=[], keepAliveMode=**DAEMON**}.`

I'm not 100% certain but I think this should be SESSION. What happens right now is that my project has multiple modules and spotbugs worker processes are left behind. When your CI runs inside containers and they have more restrictive memory limits than developer machines the containers will just get OOMkilled. Problem gets worse when you have more modules to use the spotbugs on. We have even more than main and test sourceSets that makes the issue even worse.

Even if its not that keepAliveMode then there is something that does not allow the process to end and they are left behind hogging memory.

KengoTODA:
To fix #416, run the java process as not Gradle worker but as an isolated process.
To avoid making the build process blocked, we run the process in a worker. From Gradle 6.0, we can inject `ExecOperations` to the  `WorkAction` instance so we can run a process in a worker thread (it was impossible because the worker cannot access a  `Project` instance that provides `#javaexec()` method).

This feature needs one consideration: If we launch a process by `ExecOperations#javaexec()`, we cannot touch its stdout and stderr. So in case of failure, it could be difficult to identify the root cause. For now, I found no way to handle it, so I want to release this feature as disabled-by-default. After we confirmed that this feature really solves the issue, I will enable this feature by default.

To reviewer:
Please squash-and-merge this PR with a commit message that has `feat: ` prefix, then it will release a minor update.

### DIFF
--- a/src/functionalTest/groovy/com/github/spotbugs/snom/ExtensionFunctionalTest.groovy
+++ b/src/functionalTest/groovy/com/github/spotbugs/snom/ExtensionFunctionalTest.groovy
@@ -261,7 +261,7 @@ spotbugs {
 
         then:
         assertEquals(TaskOutcome.SUCCESS, result.task(":spotbugsMain").outcome)
-        assertTrue(result.output.contains("SpotBugs 4.0.0-beta4"))
+        assertTrue(result.output.contains("spotbugs-4.0.0-beta4.jar"))
     }
 
     def "can use toolVersion to get the SpotBugs version"() {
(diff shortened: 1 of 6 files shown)


KengoTODA:
#430 will fix the broken build. See  https://github.blog/changelog/2021-01-21-github-actions-short-sha-deprecation/ for detail.

ThrawnCA (approved):

PR #435, opened 2021-02-21
fix: support for Java 8 was dropped unexpectedly

### LINKED ISSUE #434: 4.6.0 can be used with Java 1.8+, 4.6.1 requires Java 11
Unexpectedly, when bumping from 4.6.0 to 4.6.1, the plugin can no longer be used with Java 1.8:

```
java.lang.UnsupportedClassVersionError: com/github/spotbugs/snom/SpotBugsPlugin has been compiled by a more recent version of the Java Runtime (class file version 55.0), this version of the Java Runtime only recognizes class file versions up to 52.0
```

KengoTODA:
This PR will fix the broken build script, to support not only Java 11+ but also Java 8.

close #434


jscancella (approved):
LGTM

PR #465, opened 2021-03-29
fix: don't depend on jcenter for android dependencies

### LINKED ISSUE #418: Stop depending on the JCenter
[jFrog announced the end of JCenter service](https://jfrog.com/blog/into-the-sunset-bintray-jcenter-gocenter-and-chartcenter/).

Currently our build script depends on JCenter to download Trove4J, so we need to find alternative solution.
https://github.com/spotbugs/spotbugs-gradle-plugin/blob/37bb2bc2eabd98fc6506dd22c24e563df06d91b3/build.gradle#L21-L22

jscancella:
fixes #418. Currently in draft because android release is still in beta.

### DIFF
--- a/build.gradle
+++ b/build.gradle
@@ -17,15 +17,14 @@ group = 'com.github.spotbugs.snom'
 repositories {
     // To download the Android Gradle Plugin
     google()
-    // To download trove4j required by the Android Gradle Plugin
-    jcenter()
+    mavenCentral()
 }
 
 ext {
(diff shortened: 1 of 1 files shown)


KengoTODA (changes requested):
Thanks! :+1:
Just run `./gradlew spotlessApply` to format this build script, then the build will pass.

KengoTODA:
resolved by #481

PR #480, opened 2021-04-26
Upgrade to Gradle 7.0

joschi:
Refs #419


KengoTODA (changes requested):
Build is failed, could you check it?

PR #523, opened 2021-07-09
Fix groovydoc

KengoTODA:
The [latest groovydoc](https://spotbugs-gradle-plugin.netlify.app/com/github/spotbugs/snom/package-summary.html) contains strange Enum named `Confidence.1`. It also lacking description of classes and fields.


KengoTODA:
It seems that it's an unexpected behavior of Groovy v3, so I've issued a ticket to the ASF Jira.
https://issues.apache.org/jira/browse/GROOVY-10162

PR #526, opened 2021-07-13
BREAKING CHANGE: Gradle Java Toolchains support for JavaExec and WorkerAPI

rpalcolea:
With the introduction of [Toolchains for JVM Projects](https://docs.gradle.org/current/userguide/toolchains.html), there might be some folks that would like to leverage such feature when executing spotbugs either via worker api / javaexec / hybrid.

This pull request adds support for it. For now, this requires explicit opt-in via 

```
spotbugs {
          useJavaToolchains = true
}
```

If necessary, this could be default in the future but it seems like a safer approach to roll out this feature.

It will be interesting to keep an eye on https://github.com/gradle/gradle/issues/16791 

This is a breaking change as Toolchain support was introduced in Gradle 6.7 and became mature in 7.x

I suspect this should be rolled out as maybe 7.+ only? thoughts?


jscancella:
LGTM except for the latest commit message not being a semantic versioning tag

rpalcolea:
Updated the commits and intentio nbehind the change 👍

jscancella (approved):

KengoTODA:
I've merged these changes to the `beta` branch manually. Thanks for your contribution!

PR #623, opened 2021-11-24
Support creating multiple reports

### LINKED ISSUE #114: Support xml and html reports at once
I know that SpotBugs currently only supports one report at a time and that there is an open improvement request to change this, but I don't really expect this to happen anytime soon.

But as far as I have seen from a quick look, the HTML reporter just gets the same XML structure the XML reporter is generating and storing to a file (with messages enabled and minimal XML disabled) and then applies the configured XSL script to get the HTML report.

Knowing this, the Gradle plugin could either "if xml and html is enabled" forward the xml setting to SpotBugs and then apply the XSL itself to the XML report and "if xml is disabled but html enabled" either imlicitly enable the XML report with the tempdir as target and do the same to get the same result in both cases or in this case forward the HTML setting to SpotBugs itself.

This way you could generate a human readable report and a machine readable report in one go without having to do the very same thing manually in your build script.

KengoTODA:
close #114

### DIFF
--- a/src/functionalTest/groovy/com/github/spotbugs/snom/ReportFunctionalTest.groovy
+++ b/src/functionalTest/groovy/com/github/spotbugs/snom/ReportFunctionalTest.groovy
@@ -19,6 +19,8 @@ import spock.lang.Specification
 
 import org.gradle.testkit.runner.GradleRunner
 
+import java.nio.file.Path
+
 import static org.gradle.testkit.runner.TaskOutcome.SUCCESS
 import static org.junit.jupiter.api.Assertions.assertEquals
 import static org.junit.jupiter.api.Assertions.assertFalse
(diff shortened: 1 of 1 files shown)


henrik242 (approved):
lgtm

jpschewe:
So happy this is in. Is there a way to enable both xml and html reports for all tasks or does each task need to be configured individually?

KengoTODA:
@jpschewe are you using Gradle plugin to configure? If so feel free to come https://github.com/spotbugs/spotbugs-gradle-plugin to create an issue or a discussion :)

KengoTODA:
oh sorry, I misunderstood that your question was posted to the spotbugs/spotbugs project.

but please consider issuing a new issue or discussion, to make the topic visible even to who do not subscribe to this PR.

PR #627, opened 2021-11-27
chore: Include SCM information in POM

zbynek:
Including SCM information in POM should allow tools like Dependabot figure out the release information. This information is processed correctly for Spotbugs itself, so I just adapted the `scm` block from https://github.com/spotbugs/spotbugs-gradle-plugin/blob/master/gradle/publish.gradle


KengoTODA (approved):
LGTM, thanks!

Issue #614, opened 2021-11-16
XML report wasn't generated


rios0rios0:
Hello guys.

Is there some issue on v5 beta 7 regarding to the reports XML?
I ran here and the report wasn't generated. Is there a new option?
My config:
```
spotbugs {
	toolVersion = spotbugsVersion
	effort = 'max'
	reportLevel = 'low'
	ignoreFailures = false
	showStackTraces = true
	showProgress = true
	//noinspection GroovyAssignabilityCheck
	excludeFilter = file("${appResourceDir}/security/spotbugs-security-exclude.xml")
}
```

KengoTODA:
It is an intentional breaking change from the [beta1 release](https://github.com/spotbugs/spotbugs-gradle-plugin/releases/tag/5.0.0-beta.1). To generate an XML report, configure it explicitly:

```groovy
spotbugsMain {
    reports {
        xml {
            enabled = true
        }
    }
}
```

rios0rios0:
> It is an intentional breaking change from the [beta1 release](https://github.com/spotbugs/spotbugs-gradle-plugin/releases/tag/5.0.0-beta.1). To generate an XML report, configure it explicitly:
> 
> ```groovy
> spotbugsMain {
>     reports {
>         xml {
>             enabled = true
>         }
>     }
> }
> ```

```
/** @deprecated */
@Deprecated
public void setEnabled(boolean b) {
    this.isEnabled.set(b);
}

/** @deprecated */
@Deprecated
public void setEnabled(Provider<Boolean> provider) {
    this.isEnabled.set(provider);
}
```
![image](https://user-images.githubusercontent.com/32651906/142237095-cdca47ab-83cb-4ef5-839e-841790430584.png)

This is marked with **@Deprecated**. What is the right way?

KengoTODA:
Try beta 8, it can use the `required` parameter instead.

PR #643, opened 2021-12-10
Run workflows even on Windows

KengoTODA:
Recently we introduced a defect that is caused only on the Windows platform (#614). We had a similar problem (#273) previously.
To avoid reproducing these problems, run workflows in PR not only the Ubuntu platform but also the Windows platform.

This PR changes the report path from URL to canonical file path, because, on the Windows platform URL and absolute path may include short paths and it makes functional test unstable.


.github/workflows/gradle.yml
```diff
@@ -51,20 +47,49 @@ jobs:
...
+        distribution: 'temurin'
+        java-version: 11
+        cache: gradle
+    - name: Build with Gradle
+      env:
+        SIGNING_KEY: ${{ secrets.SIGNING_KEY }}
+        SIGNING_PASSWORD: ${{ secrets.SIGNING_PASSWORD }}
+      run: |
+        ./gradlew build --no-daemon "-Dsnom.test.functional.gradle=${{ matrix.gradle }}"
+    - uses: actions/upload-artifact@v2.2.4
```
KengoTODA, line 91: This version is necessary to avoid a known problem in the latest release.
Refs https://github.com/actions/upload-artifact/issues/281

PR #736, opened 2022-05-17
fix: SpotBugsTask should be cacheable even when no report is configured by user

### LINKED ISSUE #630: Regression: tasks are not cached in build cache (v5.0.0)
With 4.8.0 only tasks that needed to run are run. With 5.0.0 with build caches enabled a simple `./gradlew build && ./gradlew build` causes spotbugs tasks to run again.

Please let me know if you need more detailled information. A downgrade to 4.8.0 fixed my issue.

KengoTODA:
Treat the text file generated by our tasks, to keep the task cacheable even when it outputs no report files.

close #630

When we merge this PR, please squash-and-merge because the commit history is little bit noisy.


KengoTODA:
there is no objection for a month, so I will merge this PR by myself.

Issue #600, opened 2021-10-28
Remove SpotBugsReport.enable and SpotBugsReport.destination


KengoTODA:
These APIs will be removed in Gradle v8. Then to support Gradle v8, we need to remove these methods defined in `SpotBugsReport`.

refs #599

PR #769, opened 2022-08-16
Update SpotBugs to work with Gradle 8+

tresat:
Gradle 8 will be dropping some methods from the `ConfigurableReport` class.  By removing the `@Override marker` from some methods and marking getters as `@ReplacedBy`, this will allow a version of SpotBugs to be built which will work with Gradle 7 or 8, and can be smoke tested (internally in the Gradle build) against either version of Gradle.

After Gradle 8.0 is released, a subsequent version of SpotBugs can then fully remove these methods and update its wrapper to use Gradle 8+.

This is an incremental step towards https://github.com/spotbugs/spotbugs-gradle-plugin/issues/600

### DIFF
--- a/src/main/groovy/com/github/spotbugs/snom/SpotBugsReport.java
+++ b/src/main/groovy/com/github/spotbugs/snom/SpotBugsReport.java
@@ -51,7 +51,6 @@ public abstract class SpotBugsReport
   public abstract String toCommandLineOption();
 
   /** @deprecated use {@link #getOutputLocation()} instead. */
-  @Override
   @Deprecated
   @ReplacedBy("getRequired")
   public File getDestination() {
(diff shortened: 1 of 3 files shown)


src/main/groovy/com/github/spotbugs/snom/SpotBugsReport.java
```diff
@@ -50,8 +51,8 @@ public SpotBugsReport(ObjectFactory objects, SpotBugsTask task) {
   public abstract String toCommandLineOption();
 
   /** @deprecated use {@link #getOutputLocation()} instead. */
-  @Override
   @Deprecated
+  @ReplacedBy("getRequired")
```
big-guy, line 55: Let's just mark these as `@Internal`, they're treated as inputs elsewhere and `ReplacedBy` is meaningless outside of the Gradle codebase.

KengoTODA (approved):
LGTM, thank you!

PR #779, opened 2022-09-02
fix: use consistent naming of reports in containers

chrisdennis:


src/main/groovy/com/github/spotbugs/snom/SpotBugsTask.groovy
```diff
@@ -328,13 +328,13 @@ abstract class SpotBugsTask extends DefaultTask implements VerificationTask {
...
                     switch (name) {
                         case "html":
-                            return objects.newInstance(SpotBugsHtmlReport.class, objects, this)
+                            return objects.newInstance(SpotBugsHtmlReport.class, name, objects, this)
                         case "xml":
-                            return objects.newInstance(SpotBugsXmlReport.class, objects, this)
+                            return objects.newInstance(SpotBugsXmlReport.class, name, objects, this)
                         case "text":
-                            return objects.newInstance(SpotBugsTextReport.class, objects, this)
+                            return objects.newInstance(SpotBugsTextReport.class, name, objects, this)
```
KengoTODA, line 335: The [constructor of `SpotBugsTextReport`](https://github.com/spotbugs/spotbugs-gradle-plugin/blob/c9fe0f951e21235553ee27dddabfb9572c0a7110/src/main/groovy/com/github/spotbugs/snom/internal/SpotBugsTextReport.java#L25) has no `String` parameter.
Could you tell me who object factory will consume the `name` added by you?

chrisdennis: The various report subtypes are named managed types: https://docs.gradle.org/current/userguide/custom_gradle_types.html#managed_types

This means Gradle will generate a subclass of the type and provide a constructor that takes the name and an implementation for the `getName()` method for us: https://docs.gradle.org/current/userguide/custom_gradle_types.html#nameddomainobjectcontainer

KengoTODA: Got it, thanks! I learned one more new thing today! :)

KengoTODA (approved):

PR #780, opened 2022-09-04
Upgrade Gradle, and migrate build scripts to Kotlin DSL

KengoTODA:
Need to wait Gradle 7.6 because we cannot use `gradleTestKit()` with `jvm-test-suite` plugin at this moment.

https://github.com/gradle/gradle/issues/19849#issuecomment-1236799541


KengoTODA:
continue as #813

PR #925, opened 2023-07-31
Fix Issue 910

KengoTODA:
close #910

PR #930, opened 2023-08-01
fix: spotbugs plugin is not loaded from v5.0.15

KengoTODA:
I merged #925 to fix a known bug, but it introduced a regression.
We could not find this regression because functional test cases were not triggered during the CI workflow. So this PR contains two kinds of fixes:

1. Make `functionalTest` task running during our CI workflow
2. Set path of spotbugs plugins properly

### DIFF
--- a/.github/workflows/gradle.yml
+++ b/.github/workflows/gradle.yml
@@ -13,7 +13,7 @@ jobs:
     runs-on: ubuntu-latest
     strategy:
       matrix:
-        gradle: ['7.0', '8.0']
+        gradle: ['7.6', '8.0']
     steps:
     - uses: actions/checkout@v3
       with:
@@ -28,7 +28,7 @@ jobs:
       with:
         node-version-file: '.nvmrc'
         cache: npm
-      if: matrix.gradle == '7.0'
+      if: matrix.gradle == '7.6'
     - name: Gradle Wrapper Validation
       uses: gradle/wrapper-validation-action@v1
     - name: Build with Gradle
@@ -46,7 +46,7 @@ jobs:
         rm -rf build/libs/*.jar
         npm ci
         npx semantic-release
-      if: matrix.gradle == '7.0'
+      if: matrix.gradle == '7.6'
       env:
         GITHUB_TOKEN: ${{ secrets.GITHUB_TOKEN }}
         SIGNING_KEY: ${{ secrets.SIGNING_KEY }}
@@ -56,7 +56,7 @@ jobs:
         if [ "$SONAR_LOGIN" != "" ]; then
           ./gradlew sonarqube -Dsonar.login=$SONAR_LOGIN --no-daemon
         fi
-      if: matrix.gradle == '7.0'
+      if: matrix.gradle == '7.6'
       env:
         GITHUB_TOKEN: ${{ secrets.GITHUB_TOKEN }}
         SONAR_LOGIN: ${{ secrets.SONAR_LOGIN }}
--- a/.github/workflows/gradle.yml
+++ b/.github/workflows/gradle.yml
@@ -13,7 +13,7 @@ jobs:
     runs-on: ubuntu-latest
     strategy:
       matrix:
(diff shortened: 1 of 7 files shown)


.github/workflows/gradle.yml
```diff
@@ -13,7 +13,7 @@ jobs:
     runs-on: ubuntu-latest
     strategy:
       matrix:
-        gradle: ['7.0', '8.0']
```
KengoTODA, line 16: Gradle `7.0` does not support FooJay plugin, so need to bump up the minimal version to `7.6.2`. This change does NOT mean that our users need to use `7.6.2` or later.

buildSrc/src/main/kotlin/com/github/spotbugs/com.github.spotbugs.test.gradle.kts
```diff
@@ -72,5 +72,6 @@ tasks {
     }
 }
 tasks.check {
+    dependsOn(tasks.named("functionalTest"))
```
KengoTODA, line 75: This change is necessary to run functional test cases during the CI workflow.

src/functionalTest/groovy/com/github/spotbugs/snom/AndroidFunctionalTest.groovy
```diff
@@ -17,11 +17,13 @@ import org.gradle.internal.impldep.com.google.common.io.Files
 import org.gradle.testkit.runner.BuildResult
 import org.gradle.testkit.runner.GradleRunner
 import org.gradle.util.GradleVersion
+import spock.lang.Ignore
 import spock.lang.Requires
 import spock.lang.Specification
 
 import static org.gradle.testkit.runner.TaskOutcome.SUCCESS
 
+@Ignore
```
KengoTODA, line 26: This functional test is not working, and we have less contribution on Android support (refs #90) so simply ignore this specification for now.

src/functionalTest/groovy/com/github/spotbugs/snom/GradleJavaToolchainsSupportFunctionalTest.groovy
```diff
@@ -56,6 +56,11 @@ public class Foo {
         System.out.println("Hello, SpotBugs!");
     }
 }
+"""
+        new File(rootDir, "settings.gradle.kts") << """
+plugins {
+    id("org.gradle.toolchains.foojay-resolver-convention") version("0.6.0")
```
KengoTODA, line 62: We need this setting to use the Gradle Java toolchain with Gradle `8.0+`.

src/main/groovy/com/github/spotbugs/snom/SpotBugsBasePlugin.java
```diff
@@ -115,20 +115,12 @@ Properties loadProperties() {
...
-            .setTransitive(true);
-    project
+    return project
         .getConfigurations()
-        .create(SpotBugsPlugin.INTERNAL_CONFIG_NAME)
-        .setDescription(
-            "configuration for the external SpotBugs plugins excluding transitive dependencies")
+        .create(SpotBugsPlugin.PLUGINS_CONFIG_NAME)
+        .setDescription("configuration for the external SpotBugs plugins")
+        .setVisible(false)
```
KengoTODA, line 122: The internal configuration (`actualSpotbugsPlugins`) is needless because the FindBugs plugin is always distributed with its dependencies so no need to add transient dependencies to `spotbugsClasspath`. See [`PluginLoader`](https://github.com/spotbugs/spotbugs/blob/3cce1f0c8e9599e60c1b35030a084435b3c17a0a/spotbugs/src/main/java/edu/umd/cs/findbugs/PluginLoader.java) for detail, it uses `Class-Path` metadata of jar file to find dependencies.

KengoTODA:
I want to release this change to fix a known regression, so let me merge this for now. There is no additional test case nor new behavior then it should be OK I believe.

PR #975, opened 2023-10-08
misc: modernize things

Goooler:
You can review this PR commit by commit.

PR #988, opened 2023-10-10
misc: ignore .idea files

Goooler:
Usually, we don't need to commit any file under .idea, some files changed on my local once and once again, be easy to mis-commit them.

Refs: https://github.com/spotbugs/spotbugs-gradle-plugin/pull/975#issuecomment-1753974908.


Goooler:
CC @hazendaz.

hazendaz (approved):

hazendaz:
Thanks @Goooler !  I 100% agree on this one.  Anytime that stuff is checked in it makes rather rash assumptions that all users use same tool versions which is never accurate.  Never mind they are auto generated ;)

PR #997, opened 2023-10-14
build: use JVM Toolchains to pin the JDK version

beatbrot:
In the current state, it is essential for the project to be built and tested with JDK 11.
Otherwise either errorprone wouldn't work or tests would fail. This constraint is currently undocumented.

With this commit, we use [Gradle JVM Toolchains](https://docs.gradle.org/current/userguide/toolchains.html)
to pin the JDK that is used to test/build the project to JDK 11. If JDK 11 is already used, nothing happens.
Otherwise, a JDK 11 is downloaded and installed on the users machine.

### DIFF
--- a/build.gradle.kts
+++ b/build.gradle.kts
@@ -13,6 +13,9 @@ plugins {
 java {
     sourceCompatibility = JavaVersion.VERSION_1_8
     targetCompatibility = JavaVersion.VERSION_1_8
+    toolchain {
+        languageVersion = JavaLanguageVersion.of(11)
+    }
 }
 
 group = "com.github.spotbugs.snom"


build.gradle.kts
```diff
@@ -13,6 +13,9 @@ plugins {
 java {
     sourceCompatibility = JavaVersion.VERSION_1_8
     targetCompatibility = JavaVersion.VERSION_1_8
+    toolchain {
+        languageVersion = JavaLanguageVersion.of(11)
```
Goooler, line 17: You bumped the Java level from 8n to 11, we can use toolchain here, and remove `sourceCompatibility` and `targetCompatibility`.

beatbrot: Yes the story is more complicated here...

You guys are already using Java 11 in your CI run. So I wanted to keep you on 11. See
https://github.com/spotbugs/spotbugs-gradle-plugin/blob/17eccf7c58573576dee682b4669f6f76d7180d26/.github/workflows/gradle.yml#L24

But in addition to that, we are not able to downgrade to 8 since errorprone only supports Java 11+. We also cannot upgrade to 17 since this would break some testcases (though we could just adapt those tests). So JDK 11 is really our only choice.

If we would adapt target/sourceCompatibility, this could break some users of your Gradle Plugin since they may still be running Gradle on JDK 8.

Since I didn't want to introduce a breaking change and also wanted to be as non-invasive as possible, I came to the current configuration with source/target-Compat at JDK 8 and toolchain at JDK 11.

Does that make sense to you? :)

Goooler: > If we would adapt target/sourceCompatibility, this could break some users of your Gradle Plugin since they may still be running Gradle on JDK 8.

That's the point, after this, it requires all consumers running on Java 11 above, we should keep the compatibility of Java 8.

beatbrot: I still fear we might have a misunderstanding here.

Previously on your CI, you also used JDK 11 to run Gradle. This also meant that JDK 11 was used to compile the plugin. But with source/targetCompat, we told the compiler to emit Java 8 classes.

Now, we basically force Gradle to compile and test with JDK 11, no matter on which JVM Gradle was started. This won't change anything on your CI, since you correctly used JDK 11 anyways. But on dev machines, this will lead to everyone executing compile and test with the same JDK, no matter what was previously installed on the system. This increases reproducability. But even with this change, we still emit Java 8 class files due to source/target compat. I checked this by hand.

Goooler: You are right. I viewed the output jar, it's still on Java 8.

```sh
./gradlew assemble
cd ./build/libs  
jar xf ./spotbugs-gradle-plugin.jar
javap -v com/github/spotbugs/snom/Confidence.class  | grep major
  major version: 52
```

beatbrot: Soo....I guess we can consider this comment resolved? :) Or do you have further remarks?

Goooler: Yeah, seems we can consider using Java 17?

beatbrot: We could consider it, yeah. Currently, JDK 17 breaks `ExtensionFunctionalTest#can use toolVersion to get the SpotBugs version`. We'd have to fix this somehow.

I could look into this. Maybe in a separate PR? Or should I do it in this PR?

Goooler: Yeah, it could be in a follow-up.

Goooler (approved):
LGTM. Although, I'm not the maintainer of this project.

hazendaz:
Github Actions won't by itself download the jdk if its not the right one so this misses the mark.  Also there is little to no real reason to do this.  The source/target flags are obsolete on top of that since java 9 and given the build must build with newer jdk regardless (too many plugins are moving), its rather mute.  I'd much rather seed the 'release' flags used replacing the legacy flags and if some specific issue with tests as noted here, use junit 5 annotations to control when tests do or do not build.

I don't know if this code base was upgraded to junit 5 as I haven't looked but I would suspect it was not.  I can take that on.  I just upgraded spotbugs in recent weeks and in fact have an outstanding PR to get that building with java 11 that removes toolchains entirely.

KengoTODA (approved):
This project already depends on `org.gradle.toolchains.foojay-resolver-convention` plugin in the `settings.gradle.kts` then I think this PR is ready to merge. Thanks!

Goooler:
This could be closed after #1003 is merged.

beatbrot:
> Github Actions won't by itself download the jdk if its not the right one so this misses the mark. Also there is little to no real reason to do this. The source/target flags are obsolete on top of that since java 9 and given the build must build with newer jdk regardless (too many plugins are moving), its rather mute. I'd much rather seed the 'release' flags used replacing the legacy flags and if some specific issue with tests as noted here, use junit 5 annotations to control when tests do or do not build.
> 
> I don't know if this code base was upgraded to junit 5 as I haven't looked but I would suspect it was not. I can take that on. I just upgraded spotbugs in recent weeks and in fact have an outstanding PR to get that building with java 11 that removes toolchains entirely.

I am sorry, but I don't really understand your comment. Maybe you could somehow rephrase it? I don't see how JUnit 5 has anything to do with this change. I am also rather positive that source/target flags still works after Java 9.

hazendaz:
Hi @beatbrot, this is what I was referring to doing https://github.com/spotbugs/spotbugs-gradle-plugin/pull/1005

Currently toolchains is in spotbugs main project but when ran on GHA under java 11 it does not attempt to get java 17, it just fails with missing jdk.  Maybe that is where the wrapper action comes into play not sure.  On that project it did have a couple ways of trying to flush out a jdk usage so I migrated the entire thing to junit 5 and used annotations for same before messing with anything around toolchains logic already there.  As this project is not junit but rather spock after looking, it had similar items to setup that do the same thing.

So while toolchains can still be a valid approach, my main point is that the tests should negotiate their needs more directly as its otherwise hidden and IMHO we should run a few different jdks not just to prove the project works but to show potential contributors they can use whatever they want within reason so we get more contributors.  If we can see that the jdks are downloaded properly on GHA that would be helpful.  Personally I've never cared much for toolchains in general and there are ways around it.

Now, the configuration checks. Java deprecated source/target.  Those are not very suitable to ensure binary compatibility at all.  They were released by 'release' that does (with direct used libraries not transients).  So I was just mentioning that is a better first attempt cleaning up the build.  Yes they still work but we should not be using deprecated stuff at some point and java 9 is so distant at this point.

hazendaz:
Ultimately his PR does nothing on the remote here.  See https://github.com/spotbugs/spotbugs-gradle-plugin/blob/master/.github/workflows/gradle.yml.  If we do not expand out to run on other jdks to start with, the toolchains is effectively doing nothing.  Yes it will work for your local and yes I've confirmed my other PR runs on java 17 without issue.  So I think we need to expand a few other things in conjunction with this so we can demonstrate out all of this on GHA.

hazendaz:
most of my points are mute, gradle doesn't support doing this correctly right now.  See https://github.com/gradle/gradle/issues/25962

beatbrot:
@hazendaz Thanks for the clarification. To me it sounds like you'd rather not see this PR merged? If yes, feel free to close it. No bad feelings :)

hazendaz:
will keep open for time being.  Think there is some other work going on to make this build across other jdks / platforms / etc.  Will see what it looks like after that is done.

hazendaz:
> This project already depends on `org.gradle.toolchains.foojay-resolver-convention` plugin in the `settings.gradle.kts` then I think this PR is ready to merge. Thanks!

Thanks glad to know something exists for that as that would be useful on spotbugs too.

PR #1005, opened 2023-10-15
[tests] Add spock ignoreIf not java 11 to tests that fail on newer jdks

hazendaz:
Be very explicit on tests compatibility.  Sure tool chains can be used but it hides direct fact these are not compatible.  So lets be very clear on these.


Goooler (approved):
Passed on JDK 17 build for now.

PR #989, opened 2023-10-12
chore: polish functional tests

Goooler:
Reuse common configurations for Gradle runner.

PR #1007, opened 2023-10-16
test: use Gradle property as test Gradle version param

Goooler:
See https://github.com/spotbugs/spotbugs-gradle-plugin/pull/989#discussion_r1360009102.

### DIFF
--- a/buildSrc/src/main/kotlin/com/github/spotbugs/com.github.spotbugs.test.gradle.kts
+++ b/buildSrc/src/main/kotlin/com/github/spotbugs/com.github.spotbugs.test.gradle.kts
@@ -43,7 +43,7 @@ testing {
                         if (testGradleVersion == "current") {
                             testGradleVersion = gradle.gradleVersion
                         }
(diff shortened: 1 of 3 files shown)


buildSrc/src/main/kotlin/com/github/spotbugs/com.github.spotbugs.test.gradle.kts
```diff
@@ -43,7 +43,7 @@ testing {
                         if (testGradleVersion == "current") {
                             testGradleVersion = gradle.gradleVersion
                         }
-                        systemProperty("snom.test.functional.gradle", testGradleVersion)
+                        systemProperty("gradleVersion", testGradleVersion)
```
Goooler, line 46: Rename this key, to distinguish between Gradle property and System property.

KengoTODA (approved):
Thanks for this PR submission, I will merge this change after the v6 stable release (or I need to resolve conflicts between master and beta ;-)

This PR introduces no change to `src/main` so I've changed its subject from `fix:` to `test:`.

PR #1069, opened 2023-12-12
Target JDK8

### LINKED ISSUE #1068: Plugin does not run on JDK8
The SpotBugs plugin ships as JDK11 bytecode and therefore does not run on JDK8. This seems like an oversight, since the plugin is written in Kotlin and does not use JDK11 features.

rschmitt:
Fixes #1068.

### DIFF
--- a/build.gradle.kts
+++ b/build.gradle.kts
@@ -1,5 +1,6 @@
 import io.gitlab.arturbosch.detekt.Detekt
 import org.jetbrains.dokka.gradle.DokkaTask
+import org.jetbrains.kotlin.gradle.tasks.KotlinCompile
 
 plugins {
     groovy
@@ -54,6 +55,12 @@ signing {
 }
 
 tasks {
+    withType<JavaCompile> {
+        options.release.set(8)
+    }
+    withType<KotlinCompile>().configureEach {
+        kotlinOptions.jvmTarget = "1.8"
+    }
     named<Detekt>("detekt") {
         reports {
             sarif.required = true


KengoTODA:
As you see, the build is broken when we apply this change. So v6 release has stopped supporting Gradle running on JVM v8. It might be possible to build Java8 classfiles but build script would be tricky.

I believe that `--release` and jvm-toolchain have made it easy to build Java8 classfiles with Java17 or later. Upgrade your JVM to 17 or 21.

rschmitt:
> As you see, the build is broken when we apply this change.

Sorry, can you clarify what is actually broken? All I see is a check for class file version 55.0 (JDK11), which is failing because the class files now use version 52.0 (JDK8):

```
  echo Verifying the java version used in class files...
  cd build/classes/kotlin/main
  javap -v com.github.spotbugs.snom.SpotBugsPlugin | grep -q 'major version: 55'
```

I assume this check is here to ensure that the class file version is not inadvertently *raised*; lowering it is a compatible change. I'd fix the check, but I'm not sure where it's defined.

rschmitt:
I pushed an additional change:

```git
diff --git .github/workflows/gradle.yml .github/workflows/gradle.yml
index eeca323..1ab3de1 100644
--- .github/workflows/gradle.yml
+++ .github/workflows/gradle.yml
@@ -36,7 +36,7 @@ jobs:
     - run: |
         echo Verifying the java version used in class files...
         cd build/classes/kotlin/main
-        javap -v com.github.spotbugs.snom.SpotBugsPlugin | grep -q 'major version: 55'
+        javap -v com.github.spotbugs.snom.SpotBugsPlugin | grep -q 'major version: 52'
     - name: Run Semantic Release
       run: |
         echo "gradle.publish.key=${{ secrets.GRADLE_PUBLISH_KEY }}" > ~/.gradle/gradle.properties
```

Can the PR be reopened so the checks can run again?

PR #1075, opened 2023-12-15
fix: Gradle plugin should require JVM 8 but require JVM 11

KengoTODA:
Our commit hook　is not enough to make commits following the [Conventional Commits](https://www.conventionalcommits.org/en/v1.0.0/), so introduce `webiny/action-conventional-commits` to verify during the workflow.

This PR also adds a commit with `fix:` prefix, so it will trigger a release for v6.0.4.

### DIFF
--- a/README.md
+++ b/README.md
@@ -175,6 +175,7 @@ You can change SpotBugs version by [the `toolVersion` property of the spotbugs e
 
 | Gradle Plugin | SpotBugs |
 |--------------:|---------:|
+|         6.0.3 |    4.8.3 |
 |         6.0.0 |    4.8.2 |
 |         5.2.5 |    4.8.2 |
 |         5.2.3 |    4.8.1 |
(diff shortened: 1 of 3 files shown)


README.md
```diff
@@ -209,7 +210,7 @@ dependencies {
 
 ## Development
 ### Setup
-* development requires java 11 or higher to be installed
```
KengoTODA, line 212: Java 11 is EOL and `.java-version` is already using 17.

renovate.json
```diff
@@ -1,6 +1,6 @@
 {
   "$schema": "https://docs.renovatebot.com/renovate-schema.json",
   "extends": [
-    "config:base"
```
KengoTODA, line 4: `config:base` has been replaced with `config:recommended` so I've updated this.
Actually this is not related to the target of this PR but it should has no side-effect.

hazendaz: Per https://github.com/search?q=repo%3Arenovatebot%2Frenovate%20config%3Abase&type=code, confirmed.

hazendaz: better one here https://github.com/renovatebot/renovate/commit/f9e3e80e0c9c6a972d847f8740de5016a2bf698a

hazendaz: not sure why they are not sending auto update on that.   They should as its rather obscure, how did you even run into this?

hazendaz (approved):

PR #1081, opened 2023-12-27
[gradle] Bump gradle to 7.6.3

hazendaz:


hazendaz:
there is an unrelated issue in GHA builds, not sure what that condition is but moving this to the latest support 7.x.

hazendaz:
fixed commit id, rebuilding.

KengoTODA (approved):

PR #1082, opened 2023-12-31
ci: avoid conflict of uploaded artifacts

KengoTODA:
#1081 failed due to a conflict of uploaded artifacts, which was introduced by #1006.
This PR will resolve this issue by adding an OS identifier to the name of uploaded artifacts.


hazendaz (approved):

hazendaz:
@KengoTODA Thanks for looking into this.

PR #1087, opened 2024-01-06
handle reports as output of SpotBugsTask

### LINKED ISSUE #1077: Reports not created for cached task
Seems like 6.x regression - 5.x works as expected.

build.gradle.kts fragment:

```kotlin
spotbugs {
    excludeFilter = file("code-quality/spotbugs-filter.xml")
}

tasks.withType<SpotBugsTask> {
    reports {
        create("xml") { required = true }
        create("html") { required = true }
        create("sarif") { required = true }
    }
}

```

gradle.properties:

```
org.gradle.caching=true
org.gradle.jvmargs=-Xmx2g
```

Reproduction:
```bash
./gradlew spotbugsMain
# build/reports/spotbugs is populated
./gradlew clean
./gradlew spotbugsMain
# build/reports/spotbugs is NOT populated

./gradlew clean
./gradlew spotbugsMain --no-build-cache
# build/reports/spotbugs is populated
```

Plugin: 6.0.2-6.0.4
Spotbugs: 4.8.2-4.8.3

KengoTODA:
close #1077

I could not find the difference between the previous Groovy implementation and the current Kotlin implementation. However, it is easy to fix the reported issue; just register the report file as output of the `SpotBugsTask`.

### DIFF
--- a/src/functionalTest/groovy/com/github/spotbugs/snom/CacheabilityFunctionalTest.groovy
+++ b/src/functionalTest/groovy/com/github/spotbugs/snom/CacheabilityFunctionalTest.groovy
@@ -19,6 +19,7 @@ import org.gradle.util.GradleVersion
 import spock.lang.IgnoreIf
 
 import java.nio.file.Files
+import java.time.Instant
 
 class CacheabilityFunctionalTest extends BaseFunctionalTest {
     /**
@@ -30,7 +31,7 @@ class CacheabilityFunctionalTest extends BaseFunctionalTest {
     })
     def 'spotbugsMain task runs with configuration cache'() {
         given:
-        initializeBuildFile(rootDir)
+        initializeBuildFile(rootDir, Instant.now())
 
         when:
         BuildResult result = gradleRunner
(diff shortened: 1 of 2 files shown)


src/functionalTest/groovy/com/github/spotbugs/snom/CacheabilityFunctionalTest.groovy
```diff
@@ -164,7 +168,7 @@ class CacheabilityFunctionalTest extends BaseFunctionalTest {
         return result.output.find('Build cache key for task \':spotbugsMain\' is .*')
     }
 
-    private static void initializeBuildFile(File buildDir) {
```
KengoTODA, line 167: This `Instant` makes Java files unique for each test case. It ensures that the first build in test cases has no build cache.

src/main/kotlin/com/github/spotbugs/snom/SpotBugsTask.kt
```diff
@@ -311,6 +311,8 @@ abstract class SpotBugsTask : DefaultTask(), VerificationTask {
                 "text" -> objects.newInstance(SpotBugsTextReport::class.java, name, objects, this)
                 "sarif" -> objects.newInstance(SpotBugsSarifReport::class.java, name, objects, this)
                 else -> throw InvalidUserDataException("$name is invalid as the report name")
+            }.also {
+                (outputs as org.gradle.api.tasks.TaskOutputs).file(it.outputLocation)
```
KengoTODA, line 315: This cast is necessary to ensure we depend on no Gradle internal API. If we have no cast here, ArchUnit will warn that this line depends on an internal API `InternalTaskPutputs.file(Object`).

hazendaz:
LGTM

hazendaz (approved):

hazendaz:
@KengoTODA How did you want this merged or can you just merge it?  With the two commits I presume you wanted fix on it to auto release but wasn't sure.

KengoTODA:
Yes we can merge this PR normally, then semantic-release will check each commit and decide to trigger a patch release (because one of them uses `fix:` prefix).

Issue #1116, opened 2024-02-04
GHA Build issues Windows


hazendaz:
It appears after upgrade from jdk 17.0.9 to 17.0.10 that builds started failing on windows when running javap.  Not sure javap run is worthwhile as gradle should be compiling code right to start with.  So one solution would be just to remove that step.

hazendaz:
Issue solved itself over time.

hazendaz:
Removed for unnecessary runs

PR #1120, opened 2024-02-07
Additionally run on macos and only run javap on linux as not really needed

hazendaz:
note: The windows update done on github no longer has PATH set.  Thus the failure.  I don't see much value in javap here.  I think gradle should be ensuring its setup correctly not some post action to confirm.  As such, rather than trying to fix the path (note it runs on macos ok with path), just make it run on linux only for that step as good enough.


hazendaz:
fixes #1116

hazendaz:
@KengoTODA going to merge so repo is back to working order.  If any concerns on the change, let me know.

PR #1133, opened 2024-03-13
build: don't ignore API checking for internal package

Goooler:
We should mark all internal APIs `internal` and let BCV check visibilities.

This reverts the change in https://github.com/spotbugs/spotbugs-gradle-plugin/pull/1064#discussion_r1418501984.

### DIFF
--- a/build.gradle.kts
+++ b/build.gradle.kts
@@ -38,13 +38,6 @@ dependencies {
 val signingKey: String? = providers.environmentVariable("SIGNING_KEY").orNull
 val signingPassword: String? = providers.environmentVariable("SIGNING_PASSWORD").orNull
 
-apiValidation {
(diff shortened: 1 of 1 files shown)


hazendaz:
@Goooler Can you explain this a bit more?  I have very little idea of what you are talking about and the original PR seemed like that was asked to be added and was done so towards end before merged.   I have no issue removing it but need to fully understand how this works exactly.  Keep in mind I'm a maven user so gradle is a mystery to me.

Goooler:
We introduced Binary Compatibility Validator to track API changes in #1064 but disable checks for interval packages, it would be better to check internal packages as well, to ensure all internal APIs are marked `internal`.

CC @KengoTODA

hazendaz (approved):

PR #1210, opened 2024-07-13
build: Drop idea folder as it should not have been checked in

hazendaz:
folder is auto generated, nothing to keep here


hazendaz:
@KengoTODA Please review, I don't think intellij files should have been checked in and gitignore states to ignore them all currently.  Given that folder is auto generated, that would conflict with users workspaces.  Removing it will as well unfortunately but think it should be removed just need confirmed.

PR #1245, opened 2024-09-07
fix: remaining configuration cache issues

### LINKED ISSUE #1096: Configuration Cache compatibilities
There are still some CC compatibility issues like
```
Could not load the value of field `factory` of `org.gradle.api.internal.FactoryNamedDomainObjectContainer` bean found in field `reports` of task `:spotbugsMain` of type `com.github.spotbugs.snom.SpotBugsTask`.	
> com.github.spotbugs.snom.SpotBugsTask$$Lambda$1164/0x0000000100a5f840
```

```
1 problem was found storing the configuration cache.	
- Task `:spotbugsMain` of type `com.github.spotbugs.snom.SpotBugsTask`: invocation of 'Task.project' at execution time is unsupported.	
  See https://docs.gradle.org/7.6.3/userguide/configuration_cache.html#config_cache:requirements:use_project_during_execution
```

We can repro it at #1095

lukasgraef:
resolves #1096

### DIFF
--- a/api/spotbugs-gradle-plugin.api
+++ b/api/spotbugs-gradle-plugin.api
@@ -108,6 +108,7 @@ public abstract class com/github/spotbugs/snom/SpotBugsTask : org/gradle/api/Def
 	public final fun getClasses ()Lorg/gradle/api/file/FileCollection;
 	public abstract fun getEffort ()Lorg/gradle/api/provider/Property;
 	public abstract fun getExcludeFilter ()Lorg/gradle/api/file/RegularFileProperty;
+	public abstract fun getExecOps ()Lorg/gradle/process/ExecOperations;
 	public abstract fun getExtraArgs ()Lorg/gradle/api/provider/ListProperty;
 	public fun getIgnoreFailures ()Z
 	public abstract fun getIncludeFilter ()Lorg/gradle/api/file/RegularFileProperty;
--- a/src/functionalTest/groovy/com/github/spotbugs/snom/BaseFunctionalTest.groovy
+++ b/src/functionalTest/groovy/com/github/spotbugs/snom/BaseFunctionalTest.groovy
@@ -29,7 +29,7 @@ abstract class BaseFunctionalTest extends Specification {
         return new TestGradleRunner()
                 .withGradleVersion(gradleVersion)
                 .withProjectDir(rootDir)
-                .withArguments('--info', '--stacktrace', '--warning-mode=fail')
+                .withArguments('--configuration-cache', '--info', '--stacktrace', '--warning-mode=fail')
                 .withTestKitDir(testKitDir)
                 .forwardOutput()
                 .withPluginClasspath()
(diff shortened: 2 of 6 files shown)


KengoTODA (dismissed):
Thanks for your contribution,
All commit messages need to follow the [Conventional Commit v1.0.0](https://www.conventionalcommits.org/en/v1.0.0/), so please append `fix:` to your commit message. For detail, you may refer to [our Setup Guideline](https://github.com/spotbugs/spotbugs-gradle-plugin/tree/6.0.22?tab=readme-ov-file#setup).

hazendaz (approved):

lukasgraef:
In Gradle 7.6.3, testcases that perform two consecutive Builds using gradle's testkit, somehow fail if both builds are using config-cache. Deactivating the config-cache for one of these consecutive runs, resolves the problem. 
Therefore, I added the "--no-configuration-cache" flag to the first gradle testkit build in three testcases.

This issue is not present in the gradle 8.1 and "current" testcases.

hazendaz:
> using gradle's testkit, somehow fail if both builds are using con

I'm updating the build from gradle 7.6.3 to 7.6.4.  Can you recheck that condition after you see master updated here?

lukasgraef:
@hazendaz 
 I tried it out locally with 7.6.4, but still no luck :(
The weird thing is that the failure is caused by a ClassNotFoundException that happens only on the second consecutive run with config-cache on:
```
Caused by: java.lang.ClassNotFoundException: com.github.spotbugs.snom.SpotBugsTask$$Lambda$689/0x0000000801225728
	at org.gradle.configurationcache.serialization.DefaultReadContext.readClass(Contexts.kt:285)
	at org.gradle.configurationcache.serialization.codecs.BeanCodec.decode(BeanCodec.kt:44)
	at org.gradle.configurationcache.serialization.CombinatorsKt$reentrant$1$decodeLoop$1.invokeSuspend(Combinators.kt:165)
	at kotlin.coroutines.jvm.internal.BaseContinuationImpl.resumeWith(ContinuationImpl.kt:33)
```

mlopatkin:
> @hazendaz I tried it out locally with 7.6.4, but still no luck :( The weird thing is that the failure is caused by a ClassNotFoundException that happens only on the second consecutive run with config-cache on:

This means that the configuration cache cannot restore a lambda stored somewhere in the task state. You need to make the lambda (likely [this one](https://github.com/lukasgraef/spotbugs-gradle-plugin/blob/8090df69acee2b2f66de90f936971b098db4b9ec/src/main/kotlin/com/github/spotbugs/snom/SpotBugsTask.kt#L311)) serializable. As of 8.0, Gradle does that for you automatically, so you should only see this failure on 7.x.

You only see the failure on the second run because this issue only manifests upon loading from the cache. Starting with 8.0 or 8.1 Gradle loads the cached state even on the first run that saved the state, so there is no strict need to run the builds in tests twice.

lukasgraef:
@mlopatkin 
Thanks a lot for the explanations! You were right, the lambda you linked broke the config-cache on Gradle 7.

@hazendaz 
Now all issues should be resolved, can start another build?

hazendaz (approved):

PR #1304, opened 2025-01-08
feat: Add an extension property to decouple tasks from check

### LINKED ISSUE #1303: Request: add an option so that spotbugs tasks aren't automatically added to the check task
In some cases we may want to run the spotbugs tasks separately from `check`. I realize that we can do this with `gradle check -x spotbugsMain` but it can be a bit cumbersome to do it this way particularly when there are multiple spotbugs tasks. Many linting/formatting plugins provide an option which, when set, prevents the plugins from automatically adding the tasks as dependencies of `check`. Would you accept a PR that does this? The default behaviour would remain unchanged.

staktrace:
The new runOnCheck extension property (defaults to true) can be set to false, which disables the dependency from the `check` task to the spotbugs tasks. This is useful in cases where consumers do not want it to run automatically as part of `gradle check` but will run it in a separate context.

Fixes #1303


hazendaz (approved):

staktrace:
@hazendaz I rebased and fixed the test failure by committing the apiDump change.

hazendaz (approved):

staktrace:
Thank you!

PR #1311, opened 2025-01-16
fix: pluginJarFiles.from() can accept a Configuration directly.

autonomousapps:
This fixes a bug where local project spotbugs plugins aren't being built when running `spotbugsMain`. Here's our project structure:
```kotlin
// module/build.gradle.kts
plugins {
  ... java, kotlin, etc...
  id("com.github.spotbugs") version "6.1.1"
}

dependencies {
  ... etc...
  spotbugsPlugin(project(":my-spotbugs-plugin))
}
```
```kotlin
// my-spotbugs-plugin/build.gradle.kts
plugins {
  ... java, kotlin, etc...
}

dependencies {
  implementation("com.github.spotbugs:spotbugs:4.9.0")
  implementation("com.github.spotbugs:spotbugs-annotations:4.9.0")
}
```
And then when I run `./gradlew module:spotbugsMain`, the task _succeeds_, despite violating the rule defined in `my-spotbugs-plugin`. If, however, I add this, then the task fails as expected:
```kotlin
// module/build.gradle.kts
tasks.withType<com.github.spotbugs.snom.SpotBugsTask>().configureEach {
  dependsOn(":my-spotbugs-plugin:jar")
}
```
which led me to this PR (because that should not be necessary). There were two issues, both of which I had to fix to resolve this issue. First, this idiom is unnecessary and seems to just not work:
```kotlin
val pluginConfiguration = project.configurations.getByName(SpotBugsPlugin.PLUGINS_CONFIG_NAME)
pluginJarFiles.from(
    project.provider { pluginConfiguration.files },
)
```
that can be simplified to
```kotlin
val pluginConfiguration = project.configurations.named(SpotBugsPlugin.PLUGINS_CONFIG_NAME)
pluginJarFiles.from(pluginConfiguration)
```
Note also that I've changed `configurations.getByName()` to `configurations.named()` to maintain laziness (the latter returns a `Provider`).

Second, I changed the task input annotation from `@Internal` to `@Classpath`. This input is not, in my opinion, internal at all, but an important part of the task input. It also just doesn't work when annotated as an internal property. I suspect some other properties are also improperly marked as internal, but I haven't attempted to confirm this.

### DIFF
--- a/src/main/kotlin/com/github/spotbugs/snom/SpotBugsTask.kt
+++ b/src/main/kotlin/com/github/spotbugs/snom/SpotBugsTask.kt
@@ -294,7 +294,7 @@ abstract class SpotBugsTask : DefaultTask(), VerificationTask {
     @get:Classpath
     abstract val pluginJarFiles: ConfigurableFileCollection
 
-    @get:Internal
+    @get:Classpath
     abstract val spotbugsClasspath: ConfigurableFileCollection
 
     @get:Nested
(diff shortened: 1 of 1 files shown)


hazendaz (approved):

hazendaz:
Ironically, the original author of this plugin was asking questions to gradle on internal annotation years ago, and someone else clarified in 2023.  I'm not very skilled at gradle, but reading your write-up and searching around, I'm inclined to agree with you on all points.

Thanks for the work here!

autonomousapps:
Thank you for reviewing and releasing so quickly! Really appreciate it.

PR #1317, opened 2025-01-26
Revert "fix(deps): update dependency com.github.spotbugs:spotbugs to …

hazendaz:
…v4.9.0 (#1309)"

This reverts commit 8f73eb2b287a6ead645db157525dca4fe477fb47.

### DIFF
--- a/build.gradle.kts
+++ b/build.gradle.kts
@@ -26,7 +26,7 @@ java {
 
 group = "com.github.spotbugs.snom"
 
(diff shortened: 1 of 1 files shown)


hazendaz:
moving spotbugs back to 4.8.6 as it should not have auto updated to require java 11 on a patch release.  Will try to find where its auto releasing and remove that logic afterwards.

